PR #8, opened 2020-04-10
[Feature] Provides ability to disable weather fetching and display

camspiers:
Personally, displaying the weather doesn't have a lot of use for me
but additionally I ran into an issue where the API you are using
ratelimited call from my IP address, I guess because it was
overfetching.

This doesn't fix the ratelimited issue, but instead at least gives
people the ability to turn the feature off.

Test Plan:

In `~/.tmux.conf` set the following:

```
set -g @plugin 'tmux-plugins/tpm'
set -g @dracula-show-weather false
set -g @plugin 'camspiers/tmux'
run -b '~/.local/share/tmux/plugins/tpm/tpm'
```

Ensure that weather module no longer loads.

In `~/.tmux.conf` set the following:

```
set -g @plugin 'tmux-plugins/tpm'
set -g @dracula-show-weather true
set -g @plugin 'camspiers/tmux'
run -b '~/.local/share/tmux/plugins/tpm/tpm'
```

Ensure that weather module does stil load.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,17 @@ All instructions can be found at [draculatheme.com/tmux](https://draculatheme.co
 
 Compatible with macOS and Linux. Tested on tmux 3.0a
 
+## Options
+
+### `@dracula-show-weather`
+
+The weather module can be disabled by adding the following line to your
+`~/.tmux.conf`:
+
+```
+set -g @dracula-show-weather false
+```
+
 ## Team
 
 This theme is maintained by the following person(s) and a bunch of [awesome contributors](https://github.com/dracula/tmux/graphs/contributors).
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -3,10 +3,22 @@
 # a tmux color scheme inspired by dracula
 # author: Dane Williams
 
+get_tmux_option() {
+  local option=$1
+  local default_value=$2
+  local option_value=$(tmux show-option -gqv "$option")
+  if [ -z $option_value ]; then
+    echo $default_value
+  else
+    echo $option_value
+  fi
+}
+
 main()
 {
   # set current directory variable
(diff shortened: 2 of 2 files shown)


danerwilliams:
@camspiers ,

This is great. Just tested it out and everything works awesome. I'm gonna go ahead and merge the request. Making the theme more modular to suit everyone's needs is ideal so I'm glad you made this PR. 

Aside from that, may I ask how long you were using it before you got request rate limited? I get the current location from ipinfo.io which allows for 1000 requests per day with out authentication. The weather is then scraped from the web based on the location data obtained. The weather and location is supposed to be updated every 20 minutes so it seems a bit odd to me that you would get rate limited.

camspiers:
It was probably a few hours before I got error messages (some json) showing up in the tmux status line. Perhaps the caching wasn't working or something? Unsure.

Thanks for the plugin btw!

danerwilliams:
hmmm that's interesting. I use it all day sometimes and don't have issues but I'll look in to at least handling the rate request limit more gracefully so it doesn't just throw a bunch of json in to the status bar which I'm sure is obnoxious.

Sicnus:
Not sure where to put this and I understand this is probably the wrong place...  but the weather is being pulled from a city that's over 100 miles away from me?  I see the code in  ~/.tmux/plugins/tmux/scripts/weather.sh  and I guess I could tweak that, but it'd be nice if I could simply put a zip code preference in my .tmux.conf file to override any magic IP location gathering bits: 

set -g @dracula-show-weather 31211

?  or something like that?  Not really sure, anyway thanks so much for the work you've done.

danerwilliams:
@Sicnus can you open a new issue for this?

Sicnus:
Yessir. and thanks!

PR #10, opened 2020-04-11
Handling the 429 response limit error

cnrmrphy:
admittedly i can't test this for the use case where it doesn't 429 for at least another 20 something hours


danerwilliams:
I just tested it out and I get Request Limit Reached even though I have not reached my request limit. Maybe look in to the regex parsing or something to make sure that it is parsing correctly when the request limit has not been reached?

danerwilliams:
still the same issue with the new commit

danerwilliams:
Just tested it out and it is working great now on my end. Going to squash and merge since there are so many commits here.

Issue #11, opened 2020-04-18
Ability to set custom site for network check or have a option to disable it


SquallATF:
Because China cannot access google.com.

danerwilliams:
Closing this issue thanks to [Add options to disable battery and network #13](https://github.com/dracula/tmux/pull/13). I am going to change the default ping from google.com to github.com since presumably anyone using this tmux extension would have access to github.com in their country.

PR #13, opened 2020-04-20
Add options to disable battery and network

### LINKED ISSUE #12: Add ability to remove battery, weather, location and smiley face
The theme is fantastic, but as of right now only weather can be disabled. Some of the stuff that comes included seems unnecessary to me, being able to select which widgets I want and which ones I don't would be greatly appreciated.

smarquez1:
Related to #11 and #12


danerwilliams:
Thanks for contributing! This is great 👍

Issue #15, opened 2020-04-24
add Celsius support


PowerSnail:
Is it possible to support Celsius for temperature?

Thanks!

danerwilliams:
Good idea. Let’s add another configuration option which allows for toggling between Fahrenheit and Celsius on the weather/temperature.

danerwilliams:
#21

Issue #16, opened 2020-04-24
Rewrite weather module for international support 


danerwilliams:
Rewrite the weather module to scrape a site which supports countries outside of US. 

Requirements:
* Continue using ipinfo.io for keyless api access to current location
* Scrape weather site (do not use an API key)
* Continue getting forecast to produce appropriate emoji
* Continue getting temperature (option of F or C)

Currently thinking the best site to scrape is wunderground.com because they have a simple url structure: wunderground.com/weather/region_code/city. It looks like in the US it takes region_code as the two letter code of your state. Otherwise, it seems to expect your country code.  

For example, Vancouver, British Columbia, Canada is:  
wunderground.com/weather/ca/vancouver  

whereas Seattle, Washington, United States is:  
wunderground.com/weather/us/wa/seattle  

I think this would work, the only challenge I see is that ipinfo.io doesn't seem to give region code, only the full region name and then the country code. However, currently I am getting the state/region code by scraping another website. Since it is only the US it might be worth just making a look up table to translate them.  
  
If anyone has any other thoughts on this let me know!

danerwilliams:
I recently discovered [wttr.in](https://github.com/chubin/wttr.in) which I think would be the perfect api for this

ethancedwards8:
I could take a crack at it.

HugoLiconV:
> I recently discovered [wttr.in](https://github.com/chubin/wttr.in) which I think would be the perfect api for this

I was able to make it work using [wttr.in](https://github.com/chubin/wttr.in)

![Screenshot from 2020-07-04 17-55-49](https://user-images.githubusercontent.com/22970186/86522738-b816ff80-be1f-11ea-85fe-79087fc104a1.png)

ethancedwards8:
Nice, beat me to it

ethancedwards8:
You should submit a pull request

danerwilliams:
Please submit a PR, also if you we should include the current location as well as the weather as it currently does.

ethancedwards8:
#52

PR #17, opened 2020-04-24
add option for celsius

PowerSnail:
For #15.

### DIFF
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -21,6 +21,7 @@ main()
   show_battery=$(get_tmux_option "@dracula-show-battery" true)
   show_network=$(get_tmux_option "@dracula-show-network" true)
   show_weather=$(get_tmux_option "@dracula-show-weather" true)
+  temperature_unit=$(get_tmux_option "@dracula-temperature-unit" "F")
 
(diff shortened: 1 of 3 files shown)


danerwilliams:
So a couple things on this:
* Could we avoid setting an environment variable to represent the unit of temperature
* Instead of changing the regex filter based on the specified unit of temperature, maybe converting from F to C with a conversion formula would be simpler?

PowerSnail:
I've changed it to use cli argument instead of environment variable.

Regarding the second point, may I ask why? Since the `curl`ed result already contains `deg;C`, what's wrong with `grep`-ing it?

danerwilliams:
So I wouldn't say there is anything wrong with this approach and it seems to work fine. There are just two reasons why I think another approach would be better:
* I plan on scraping a different site to add international support for weather as mentioned in #16 
    - If we just made a simple conversion function to celsius then this code would be more reusable
* By setting an option with "f" or "c" I feel like there is more room for error by the user by passing in a wrong input to that option
    - If we just make it something like weather-celsius false (or true) then everything remains pretty much fool proof (I know this is nitpicking to an extent)

danerwilliams:
#21

PR #18, opened 2020-04-25
(Sorry I did not notice this issue has already been attempted by others) Update weather.sh

yizeqi:


danerwilliams:
A couple of things here:
* There needs to be a configuration option to toggle between F and C (like how you can turn off or on certain modules)
* I think there are some typos here and also some incorrect variable references so make sure you test your code

yizeqi:
Thank you for the feedback! Sorry that I didn't understand the prompt that clearly.

PR #20, opened 2020-04-29
scripts/network.sh: use a list of hosts

pbui:
- Use a list of hosts: google.com, github.com, example.com
- Script will try each in turn, stopping at the first one that connects
- Also fix she-bang and some quotes

- Tested with Wifi and Offline.


danerwilliams:
looks good to me!

PR #22, opened 2020-05-11
Added powerline symbols

uttarayan21:
Changed it so that battery can be read from BAT1 and BAT0

### DIFF
--- a/scripts/battery.sh
+++ b/scripts/battery.sh
@@ -1,11 +1,13 @@
 #!/usr/bin/env bash
 
+BAT=$(ls /sys/class/power_supply/BAT* | head -1)
+
 battery_percent()
 {
 	# Check OS
 	case $(uname -s) in
 		Linux)
-			cat /sys/class/power_supply/BAT0/capacity
+			cat $BAT/capacity
 		;;
 
 		Darwin)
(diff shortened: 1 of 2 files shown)


danerwilliams:
This is great! Before I merge could you make it an option for the user to turn powerline symbols on or off?  

i.e.  `set -g @dracula-show-powerline false`

uttarayan21:
Hi I added the flag to show the powerline symbols. It's a bit hacky but it works

danerwilliams:
Thanks for adding the configuration. Only other things I think we need to merge this are:
* for the far left smily face make it so that in non powerline mode it doesn't add an extra space on the right side
* the arrows for the window tabs seem to not be quite right. For example in window 0 the arrows are pointing towards each other instead of both of them to the right

Screen Shot 2020-05-12 at 1 39 12 PM

uttarayan21:
Hi thanks for the suggestions I added the features as you asked. However in the window zero the gap is put delibeately. Since there is a powerline glyph so is'nt completely flush with the left part so it looks a bit weird imo.
Screenshot from 2020-05-15 11-18-43
But If you want it like that I can remove the space.

danerwilliams:
This is great! I don't have too much preference one way or the other especially as I'm not a powerline user myself. If powerline people disagree on the space then it wouldn't be hard to remove later. I'm gonna go ahead and merge the request. Thanks for contributing!

Also sweet setup btw.

PR #23, opened 2020-05-19
Adding military time functionality

ethancedwards8:
I think I've did this correct, if I haven't please let me know and I'll be happy to fix this

The way it works is you can set `set -g @dracula-tmux-miiltary true` and it'll set the time in the bottom to military time instead of a 12 hour time. This is my first PR so if I did anything wrong my bad.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -20,3 +20,4 @@ Customize the status bar by adding any of these lines to your .tmux.conf as desi
 * Disable weather functionality: `set -g @dracula-show-weather false`
 * Switch from default fahrenheit to celsius: `set -g @dracula-show-fahrenheit false`
 * Enable powerline symbols: `set -g @dracula-show-powerline true`
+* Enable military time: `set -g @dracula-military-time true`
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -20,10 +20,7 @@ main()
   show_network=$(get_tmux_option "@dracula-show-network" true)
   show_weather=$(get_tmux_option "@dracula-show-weather" true)
   show_fahrenheit=$(get_tmux_option "@dracula-show-fahrenheit" true)
-<<<<<<< HEAD
-=======
   show_powerline=$(get_tmux_option "@dracula-show-powerline" false)
->>>>>>> 4084fd83c8f5b4a71eb1be8e9421c1a68cff4c3d
   show_military=$(get_tmux_option "@dracula-military-time" false)
 
   # Dracula Color Pallette
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -48,6 +48,8 @@ main()
   # tmux set-option -g clock-mode-style 12
   if $show_military; then
     tmux set-option -g clock-mode-style 24 &
+else
+    tmux set-option -g clock-mode-style 12 &
   fi
 
   # set length 
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -49,12 +49,7 @@ main()
   tmux set-option -g status-interval 5
 
   # set clock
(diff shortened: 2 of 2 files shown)


danerwilliams:
I tried testing it and it didn't seem to work. It also looks like you deleted the parameter for showing powerline by accident. Can you make changes to fix and then test to see if it works on your end then push changes? Otherwise idk why it's not working on my end.

ethancedwards8:
The reason I'm missing the powerline stuff is that all of this is from a outdated branch, I originally cloned the repo like a few weeks ago, because I had that idea then, but I just had the time to do it now, I'll try and fix whatever happened though. If you could give me some guidance on how to get it back up to date that'd be great.

danerwilliams:
oh ok that makes sense. It's not very difficult for me to resolve those merge conflicts so don't worry about that. When I resolved them on my local though military time didn't work when I set my configuration. Is this consistent with what's happening on your end?

ethancedwards8:
I actually just figured it out and its synced now lol. But yes, oddly enough it was working earlier, let me see if I accidently deleted any code, or if there's something I've overlooked. Do you notice any immediate errors in the code?

danerwilliams:
Honestly everything made sense to me so I was a bit confused why It wasn't working lol, I'll try again in a bit, may have just been something weird on my end.

ethancedwards8:
Well oddly enough it isn't working for me now, I did regenerate my local (I'm running Linux) earlier today, maybe that's causing it, should I make a new file dedicated to that function like how the other ones are?

danerwilliams:
I'm not sure what you mean by that, what are the other one's you're referencing?

ethancedwards8:
Like if I want to disable the Fahrenheit converter on the weather script, if you go into `dracula.sh` and go down to the `if $show_weather; then` it runs a script from another file, I'm saying do you think separating the script into a different file would fix the issue?

danerwilliams:
Oh no that is definitely not necessary, it was originally set by just just that one line of code that you put if statements around.

I may be totally off on this, but I think I remember that 24 hour was the default when I originally made this project and I had to change it to 12 hour by adding that line. Maybe try researching a little bit more or seeing what happens when you just get rid of that part and have nothing there? ie eliminate the 12 and 24 hours things you made and see what happens, if it does in fact default to 24 then just add an if statement so that if show_military is false then change to 12 hour.

ethancedwards8:
Oh hmm, that may totally be a possibility, let me try that, I have a feeling you may be correct in that.

danerwilliams:
Also, once you have it working please add to INSTALL.md under the configuration section.

ethancedwards8:
So this may be a stupid question but, I can't seem to get tmux to reload correctly, I'm doing `tmux source ~/.tmux.conf` and then doing `prefix + I` but nothing is changing, am I doing this right? Also yes, I can add to that, I was going to do that originally in my PR but I forgot to.

danerwilliams:
I've found that method isn't very effective for testing out code on tmux, it usually isn't responsive for me either. It also requires that you have changes committed and pushed which is not ideal. 

I would just run the dracula.tmux executable in whatever local repo you are editing on: `./dracula.tmux`

ethancedwards8:
Ohhh okay thank you, that makes sense.

ethancedwards8:
I think I may have figured the issue out, tmux uses a time system called strftime, and in some of the code, its setting it to use standard 12 hour time, I think I know a way to fix this, but I'll have to work on it tomorrow, it's getting quite late where I am. Thanks for all the help tonight, I'll try and get you a working commit tomorrow.

ethancedwards8:
Everything should be working! Working perfectly on my end, learned a bit more about git today, my suspicion last night was correct.

Issue #14, opened 2020-04-21
Change smiley face to current session on left side of status bar


danerwilliams:
Related to discussion with @smarquez1 in [Add ability to remove battery, weather, location and smiley face #12](https://github.com/dracula/tmux/issues/12).
  
Change the smiley face on the left side of the status bar to have 3 options:
* Current smiley face (changes color based on prefix activation)
* Nothing at all
* Current session name
  
I'm curious if people have other ideas for this side of the status bar as well!

aquateen:
I would love to have the session name in the status bar 🙏

Lulus792:
set -g @dracula-show-left-icon-session

PR #24, opened 2020-05-21
Added option to change the smiley face

uttarayan21:
rewrote a function in battery.sh to make use of acpi command in linux if it exists or else fallback to /sys/class/power_supply/BAT*
added option to switch powerline separators
added option to switch the smiley face with window name or any other character issue #14

### DIFF
--- a/scripts/battery.sh
+++ b/scripts/battery.sh
@@ -1,13 +1,39 @@
 #!/usr/bin/env bash
+linux_acpi() {
+    arg=$1
+    BAT=$(ls -d /sys/class/power_supply/BAT* | head -1)
+    if [ ! -x "$(which acpi 2> /dev/null)" ];then
+        case "$arg" in
+            status)
+                cat $BAT/status
+            ;;
 
-BAT=$(ls /sys/class/power_supply/BAT* | head -1)
+            percent)
+                cat $BAT/capacity
+            ;;
 
+            *)
+            ;;
+        esac
+    else
+        case "$arg" in
+            status)
+                acpi | cut -d: -f2- | cut -d, -f1 | tr -d ' '
+            ;;
+            percent)
+                acpi | cut -d: -f2- | cut -d, -f2 | tr -d '% '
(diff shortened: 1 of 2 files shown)


uttarayan21:
Screenshot from 2020-05-21 21-40-17

Here is a screenshot of how it looks now

ethancedwards8:
Dang, I was going to do this today, beat me to it :+1: :smile:

ethancedwards8:
Also, you might want to make an entry in INSTALL.md

danerwilliams:
Some users have reported issues with the change in battery.sh from #22 so it was reverted in back to the original battery script with #26 . 

Can you resolve merge conflicts and use the fall back option to what was merged in #26 ? 

Also, can you explain the advantage of using acpi command if it exists?

uttarayan21:
Oh sorry about the #22 commit it didn't test it extensively.
And about the acpi command, it works well for me while the /sys/class/power_supply/BAT0 usually changes in between reboots and I have to change some of my scripts.
Also right now I'm stuck in a power outage due to a cyclone amphan and won't be able to change the commits maybe for a few days.

uttarayan21:
Hi I fixed the merge conflicts and added entries to INSTALL.md

danerwilliams:
The changes to dracula.sh are great. 

My only concern is with the changes to battery.sh. In #26 the battery was fixed by reverting back to it's original state, and it looks like for the fall back if acpi isn't present that the fallback is what battery.sh looked like before it was fixed in #26 . Can you please change it accordingly?

uttarayan21:
Yes the changes before #26 were bugged but now its a llittle different
before #26 I tried to list all the folders in /sys/class/power_supply/ starting with BAT
but that led to ls listing all files inside BAT0 or BAT1 so I've done this
ls -d /sys/class/power_supply/BAT* 
which only lists directories in /sys/class/power_supply that start with BAT
I hope this addressed your concerns.

PR #30, opened 2020-05-27
Add option to enable high contrast on pane borders

ImperadorSid:
Sometimes, when several panels are open, it is difficult to know which panel is active due to the low contrast of the _dark_purple_ color in the _purple_ background.
So I implemented a simple option that changes the color to _light_purple_ when the `@dracula-border-contrast` flag is `true` (the default is `false` when the flag has not been set).


ImperadorSid:
Thanks!
This was my first pull request.

Issue #36, opened 2020-05-30
Enhancement: display CPU & memory usage


ciehanski:
### Describe the feature you'd like
It'd be nice to have dracula tmux include the option to enable the display of CPU & memory usage meters within the statusline similar to [tmux-mem-cpu-load](https://github.com/thewtex/tmux-mem-cpu-load).

### Describe risks you've considered
I'm really unsure.

### Describe alternatives you've considered
I've tried installing [tmux-mem-cpu-load](https://github.com/thewtex/tmux-mem-cpu-load), but because I am using dracula theme, it will not display.

ethancedwards8:
I'm taking a crack at it.

ethancedwards8:
I solved it and made a PR :)

ciehanski:
@ethancedwards8 wow, amazing work!! I really appreciate it, this will be incredibly useful.

ethancedwards8:
I just added RAM percentage support!

backpropper:
I'm aware of the above repository but the functionalities I mentioned above have not been implemented correctly in any plugin. The closest it comes to is the [https://github.com/tmux-plugins/tmux-cpu](https://github.com/tmux-plugins/tmux-cpu) repo which does not handle this. 

Nonetheless, this kind of functionality is super useful for thousands of developers who use remote GPU servers such as in AWS, GCP Azure, etc. Hence having this added functionality in addition to the existing CPU/RAM usage tool would make it a powerful plugin. I know many people migrated to the Dracula theme not only because of the looks but also because of the current functionality of CPU usage is seamlessly integrated into the theme.

ethancedwards8:
I still have the code for the GPU module, I can start working working on it again. I had just stopped because @danerwilliams said we need a way to implement it a different way. But I do have that code.

danerwilliams:
@backpropper @ethancedwards8 I’m in favor of trying to bring that functionality then. 

Really appreciate your feedback.

ethancedwards8:
So keep working on the code? Or not? I'm caught up in some work for my internship, so I can't work on the custom listing, but I can try finish up the GPU, and the RAM modules.

danerwilliams:
It sounds like there is a significant demand for the functionality @backpropper mentioned, so I would plan on merging that feature if you finished it. 

You should probably open a new issue for it since it’s a bit different than RAM/CPU.

ethancedwards8:
Okay, will do.

PR #41, opened 2020-06-04
Added RAM Percentage per issue #36

ethancedwards8:
Working cleanly on Linux powerline and non powerline


ethancedwards8:
#36

danerwilliams:
So now that we have a lot more information to potentially be displayed in the status bar I think we need to consider what colors we should be using or determining a way for the colors and / or the order of the different modules are displayed on the status bar. What is everyone's thoughts? It just looks kind of weird to me when we have two orange modules and when this is merged two cyan modules.

ethancedwards8:
I agree, I didn't know what colors to choose, I was just trying to get the functionality, since you're the owner I was thinking you could just pick the colors.

ethancedwards8:
I was thinking about a way to choose what order the modules go in, maybe using an array, and you can specify the order in the array, within the .tmux.conf file

ethancedwards8:
I was also thinking about combining the RAM and CPU modules, and if both are enabled in the .tmux.conf file then it could be in one segment in the bar, but if only one is enabled then its still a singular segment, so if both were enabled it would be the same color, and closer together, if that makes any sense

danerwilliams:
Yeah I definitely think we need to make it more configurable somehow. Having a list in the configuration like you mentioned could be a good idea. I'll merge for now to give the functionality and then raise an issue for making it more configurable.

danerwilliams:
Before I merge are you able to add mac compatibility?

ethancedwards8:
I do have a command I think will work, but I can't test it currently

ethancedwards8:
`ps -A -o %mem | awk '{mem += $2} END {print mem}'`

ethancedwards8:
I think that will work, if you could test it and give me the output that'd be great

ethancedwards8:
Man, I really need to get a Mac lol, I borrow my moms whenever I need to test code.

ethancedwards8:
Also another interesting question, should we add support for the *BSD's?

ciehanski:
@ethancedwards8 looks like `ps -A -o %mem | awk '{mem += $1} END {print mem}'` works for me on my mac, but just to specify, it only polls _app_ memory and not _wired_ memory as well.

ethancedwards8:
Okay, forgive the stupid question but, what’s the difference between app memory and wired memory? I could probably achieve the total_used/total_mem, but the reason I chose the percentage is because the CPU is percentage based.

danerwilliams:
So what I’m reading is that the total memory used is the sum of wired memory, app memory, and compressed memory. Maybe there is a way to get memory used with out needing to individually gather all 3 of those figures?

danerwilliams:
And if you would like support for BSD, then please open an issue.

ethancedwards8:
I think we should just support what we're doing right now, and if someone comes along and wants BSD support I think we could do it. But for the memory issue, I'm not sure, if you run the `free -m` or `free -h` command what does it output?

ethancedwards8:
Actually, I think that's a Linux only command, I'll search through stackoverflow and see if I can find something.

ciehanski:
Look into:
```bash
❯ top -l 1 -s 0 | grep PhysMem
PhysMem: 14G used (3148M wired), 2503M unused.
```
You could probably grep the output and get the actual results.

ethancedwards8:
Interesting.. thanks for finding that, did you find this off of anything? Or did you come up with this, if you found it, can you link the source please? I feel this could be what we are looking for, sadly it doesn't run on Linux, so it'll be hard for me to test, so I'll be in touch.

ciehanski:
I found it [here](https://apple.stackexchange.com/a/46655), good ol' stackx. I'll play around with the command and try to get something that works, but I am willing to help test any command you come up with. I am going with what the Activity Monitor reports which is `Memory Used` as `PhysMem(used) - wired(used)`.

Final output in dracula should look like (IMO):
 ```bash
10G/16G
```

ethancedwards8:
Okay, well I'm on vacation, so I probably won't be able to get out a solution for about a week. So, if @danerwilliams wants to merge my current PR, or leave out a commit or whatever he wants to do, you can take over and do the Mac support, and I can try and help out when my vacation is over.

ethancedwards8:
Does this work? I had some free time and decided to work on this, if this works tell me the output, if it doesn't, back to the drawing board I guess.

ethancedwards8:
[How it looks](https://drive.google.com/file/d/1GaoV1LeoP_62JHDl8qyxlQ5A6_m2XD4E/view)

ethancedwards8:
Okay, sorry for all of the excess commits, I was testing on my moms old mac, and I was pushing and editing the code from my desktop.

ethancedwards8:
Everything is working, I renamed the variables from 'dracula-***-percentage' to 'dracula-***-usage', because for the RAM it shows the usage/total per @ciehanski request, thats not the percentage thats the usage.

ethancedwards8:
Sooo any feedback?

danerwilliams:
Just tested and it seems like RAM on mac isn't accurate. it's saying 13G/16GB when htop is showing 7 gb usage. Also, why just G for the used and GB for available?

ethancedwards8:
The G instead of GB was unintentional, I didn't see it.

ethancedwards8:
I've been parsing ram out from top, and it wasn't the same as htop, I've yet to find a command that gives me the same output to htop. I also tried vm_stat

danerwilliams:
hmmm, it's weird to me that all of these different programs would be getting different readings

danerwilliams:
From this it sounds like you maybe could get the same number as htop with the data provided in top if you do some calculations:
https://askubuntu.com/questions/791270/why-are-memory-statistics-in-top-and-htop-different

ethancedwards8:
Okay, I'll look into that, thanks.

ethancedwards8:
Okay, I had some spare time today and was working on a solution. I should have a working commit tomorrow or the day after that.

ethancedwards8:
Everything is good now, if it is higher than htop the reason behind that is it rounds up.

PR #44, opened 2020-06-17
Makes option to set weather to always display 69 degrees

billyporter:


billyporter:
This is not what opensource was built for

Issue #45, opened 2020-06-18
Enhancement: GPU usage 


ethancedwards8:
### Describe the feature you'd like
GPU usage, in the bar

### Describe risks you've considered
None

### Describe alternatives you've considered
I have some code that works, I just haven't fixed all the issues yet, I will finish it when I have some time.

ethancedwards8:
Worked on this last night, should have a PR tonight.

PR #50, opened 2020-06-30
Adding GPU usage.

ethancedwards8:
#45 @backpropper

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -24,5 +24,6 @@ Customize the status bar by adding any of these lines to your .tmux.conf as desi
 * Enable military time: `set -g @dracula-military-time true`
 * Switch the left smiley icon `set -g @dracula-show-left-icon session` it can accept `session`, `smiley`, `window`, or any character.
 * Enable high contrast pane border: `set -g @dracula-border-contrast true`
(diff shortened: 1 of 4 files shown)


ethancedwards8:
I have an nvidia on my system, so I don't have any way to test an AMD gpu, also I don't have access to a Mac with a GPU so I'm stuck there. This is just getting the ball rolling.

danerwilliams:
Is there no way to monitor gpu's in general?

danerwilliams:
It looks like we can see how they calculate it in this [tmux-gpu plugin](https://github.com/tmux-plugins/tmux-cpu). Appears they only have nvidia compatibility as well.

ethancedwards8:
I'm not aware of a way to monitor GPU's in general.

danerwilliams:
I don't have a nvidia gpu to test myself on, but since it can be disabled and it seems like this is consistent with other implementations of the gpu usage percent, I am going to trust your tests and merge it.

backpropper:
Thanks for adding this @ethancedwards8. I haven't seen any API for checking gpu usage for AMD/Intel GPUs so might as well ignore for now.

backpropper:
It seems like now the status bar does not show any of the CPU/GPU percentage? Did anything change?

Issue #42, opened 2020-06-04
Enhancement: Make status bar modules / color more configurable


danerwilliams:
### Describe the feature you'd like

Instead of just having enable / disable for the various modules (cpu, weather, battery, etc.) have a configuration option which accepts a list specifying the desired modules in the desired order for them to appear. There should also be a configuration option which specifies the corresponding dracula colors for these modules. 

### Describe risks you've considered

The user may not set this configuration option

### Describe alternatives you've considered

If not specified, default to make dracula theme look like it does in the screenshot

ethancedwards8:
I have an idea for how we would implement it. When we have the user specify the order of the modules in the .tmux.conf, we can store it as an array, then loop through it and display it that way. For the colors we will do the same thing. Just an idea.

danerwilliams:
Yes I may have been unclear about this but that’s is what I meant by accepting a list as a configuration options.

PR #52, opened 2020-07-06
add international support for weather module

HugoLiconV:
It uses [wttr.in](https://github.com/chubin/wttr.in) to get the weather information and the current city based on the IP address

Solves issue #16

### DIFF
--- a/scripts/weather.sh
+++ b/scripts/weather.sh
@@ -27,8 +27,8 @@ display_weather()
 	fi
 	weather_information=$(fetch_weather_information $display_weather)
 
-	weather_condition=$(echo $weather_information | awk '{print $1;}') # Sunny, Snow, etc
-	temperature=$(echo $weather_information | awk '{print $2;}') # +31°C, -3°F, etc
+	weather_condition=$(echo $weather_information |  cut -d "+" -f 1 | cut -d "-" -f 1) # Sunny, Snow, etc
+	temperature=$(echo $weather_information | cut -d '+' -f 2) # +31°C, -3°F, etc
 	unicode=$(forecast_unicode $weather_condition)
 
 	echo "$unicode ${temperature/+/}" # remove the plus sign to the temperature
@@ -36,13 +36,13 @@ display_weather()
 
 forecast_unicode() 
 {
-	weather_condition=$1
+	weather_condition=$(echo $weather_condition | awk '{print tolower($0)}')
 
-	if [[ $weather_condition =~ 'Snow' ]]; then
+	if [[ $weather_condition =~ 'snow' ]]; then
 		echo '❄ '
-	elif [[ (($weather_condition =~ 'Rain') || ($weather_condition =~ 'Shower')) ]]; then
+	elif [[ (($weather_condition =~ 'rain') || ($weather_condition =~ 'shower')) ]]; then
 		echo '☂ '
-	elif [[ (($weather_condition =~ 'Overcast') || ($weather_condition =~ 'Cloud')) ]]; then
+	elif [[ (($weather_condition =~ 'overcast') || ($weather_condition =~ 'cloud')) ]]; then
 		echo '☁ '
 	elif [[ $weather_condition = 'NA' ]]; then
 		echo ''
--- a/scripts/weather.sh
+++ b/scripts/weather.sh
@@ -3,79 +3,29 @@
 
 fahrenheit=$1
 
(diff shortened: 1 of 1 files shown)


ethancedwards8:
Nice, it's working for me

ethancedwards8:
Is ipinfo a global api? Also just my two cents, I think we should try and avoid using two external api’s

scripts/weather.sh
```diff
@@ -18,14 +25,38 @@ display_weather()
 	else
 		display_weather='&m' # for metric system
 	fi
-	echo $(weather_information $display_weather)
+	weather_information=$(fetch_weather_information $display_weather)
+
+	weather_condition=$(echo $weather_information | awk '{print $1;}') # Sunny, Snow, etc
```
HugoLiconV, line 30: I'm splitting the string in two to get the temperature and the weather condition, the latter is used to get the Unicode character. 
**wttr.in** returns something like this: `Sunny 36°C`, so the weather condition is `Sunny` and the temperature is `36°C`. 
For splitting the word I'm using `awk` but there's another approach transforming the word into an array and then accessing the words  using `array[0]` and `array[1]` but,  I don't know which of these is the best approach,

danerwilliams: why not use cut? i.e. `echo "Sunny 36°C" | cut -d ' ' -f 1` returns `Sunny`

HugoLiconV: I'm quite new to bash so I don't know a lot of tools, but I'll change it to `cut`. I just found two bugs, wttr.in sometimes can return this: **"Partly cloudy +34°C"**. The first bug is that we split the string using a space as a delimiter, this won't work because when we try to get the temperature, instead of getting **+34°C** we will get **cloudy**. So, I was thinking of 
 doing this: 
```bash
# Split the word using the plus or minus sign
weather_condition=$(echo "Partly cloudy -34°C" | cut -d "+" -f 1 | cut -d "-" -f 1) 
temperature=$(echo "Partly cloudy -34°C" | cut -d "+" -f 2 | cut -d "-" -f 2)
```
Here  I'm using `cut` twice because I don't know a way of specifying two delimiters. So if you know a better approach just let me know.  
The second bug is when we try to get the Unicode character. If we use the same example from above (**"Partly cloudy +34°C"**), then this case won't work 
```bash
elif [[ (($weather_condition =~ 'Overcast') || ($weather_condition =~ 'Cloud')) ]]; then
```
because **cloudy** is in lowercase. Here I was thinking of transforming the variable `weather_condition` to lowercase, what do you think?

danerwilliams: Yes, I think that could be a good idea. Obviously the keywords are going to be a bit different because the verbiage used is different on wttr.in than the previous weather data site that was being scraped. Implement whatever works / makes the most sense to you.

ethancedwards8:
@HugoLiconV what's the status on this? I would like to get this merged for you but it needs to be fully working :)

Also, minor issue that we can solve with #42 is that currently the CPU and Weather module have a color conflict. Nothing wrong on your part we just need something to fix #42 :)

HugoLiconV:
> for

Hi @ethancedwards8, the code should be working now, I did some changes but I didn't receive a response or review from @danerwilliams

danerwilliams:
@HugoLiconV sorry I think I have changes requested still so that is why it hadn't been merged

HugoLiconV:
> @HugoLiconV sorry I think I have changes requested still so that is why it hadn't been merged

I updated the PR after you requested the changes, could you tell me what it's missing?

danerwilliams:
Oh I'm not sure why but the change request didn't resolve. Either @ethancedwards8 or I need to test and confirm the change but otherwise it seems fine.

ethancedwards8:
It's working well for me. Since this is such a big change I think @danerwilliams should also test it out, then we can merge it if there's no issues. Just in case.

ethancedwards8:
I've been using it the past 2 days and it's worked great. Merging! :)

sheikki:
In general, I use tmux on VPN'd connections on remote servers/clusters. As such, the IP-tied weather info isn't very relevant to me. I would like it to be though..

kedare:
Is there a way to force the location ? The IP addressing plan of my ISP is not location based the national-wide so I get the weather from Paris instead of my city.

ypicard:
I agree, forcing the location is needed when using a VPN.

Issue #48, opened 2020-06-26
Enhancement: [add this functionality] Uk date format


bashleigh:
### Describe the feature you'd like
would it be possible to add an option to change the date format or even use locale? Will look to see if I can add it myself but I'll be honest, haven't used shell much!

ethancedwards8:
#62

PR #62, opened 2020-08-25
Add flag for dd/mm date format that works with military time.

playonverbs:
This adds a setting to allow changing the date `mm/dd` format on the right status to a `dd/mm` format on as used in countries like the UK. This works in tandem with the military time setting. I've tested only the non-powerline version thanks to terminal limitations but should cover powerline users too.


ethancedwards8:
#48

ethancedwards8:
@playonverbs can you fix the merge conflicts with the master branch? Otherwise it looks good to me

playonverbs:
@ethancedwards8 Thanks! The merge conflicts have now been resolved

ethancedwards8:
Thanks!

PR #63, opened 2020-08-31
Specify numeric locale for parsing top output

gaganpreet:
My locale is set to `nl_NL.UTF-8`, and it uses comma as a decimal separator and breaks the parsing of top output.

This change uses US locale which will generate output with point as a decimal separator.


ethancedwards8:
Looks good.

Issue #70, opened 2020-09-12
Bug: git pull failed due to weather.txt chnages


zethra:
### Describe the bug
The sleep_weather script writes to weather.txt which is tracked by git.  This file being changed means that git pulls fail due to upstaged changes (at least on my system, I have the default pull stagey as rebase).  Either way, have the script write to a checked in file isn't a good idea.

### To Reproduce
1. Run tmux with with the weather plugin enabled
    1.5 (might be needed) Set the repo or global config to rebase on pull instead of merge
2. Try to run git pull on the plugin

### Expected behavior
The plugin shouldn't write to checked in files.

### System
 - OS: Linux (Manjaro)
- Tmux Version: 3.1b

zethra:
I submitted a patch to fix this https://github.com/dracula/tmux/issues/70

ethancedwards8:
Closed by #71

PR #71, opened 2020-09-12
Fix bug #70

zethra:
Changed weather_sleep.sh and dracula.sh so that they do
not assume that weather.txt exist.  As a result I was
able to remove weather.txt as a tracked file.


zethra:
Fixes https://github.com/dracula/tmux/issues/70

ethancedwards8:
Okay nice, I don't use the weather module so I've never encountered this problem, LGTM.

PR #58, opened 2020-08-06
Support multi-gpu environment

jaywonchung:
Did the following:
1. Support homogeneous-vendor multi-GPU machines. Utilization values are simply averaged.
2. Grepping with `VGA` is more general.
3. Print `GPU unknown` for other GPU vendors.

Tested on my desktop (one NVIDIA GPU, Ubuntu bionic), and my lab server (six NVIDIA GPUs, Ubuntu bionic).

PR #72, opened 2020-09-16
Show all GPU stats in multi-GPU environment

jaywonchung:
**Context**
- Currently, in multi-GPU environments, Dracula shows the average utilization of all GPUs (I did that in #58). Yet, I believe it is better to show the utilization of all GPUs.

**Major Changes**
- For multi-GPU environments, refer to the following screenshot. Things are the same for single-GPU environments.

스크린샷 2020-09-17 오전 12 04 52

**Comments**
- I fixed all utilization percents to show two digits so that the length of the plugin bar doesn't fluctuate. This means that if the utilization is under 10%, it will be prefixed by one zero.

### DIFF
--- a/scripts/gpu_usage.sh
+++ b/scripts/gpu_usage.sh
@@ -32,8 +32,13 @@ get_platform()
 get_gpu()
 {
 	gpu=$(get_platform)
+	expand=$(get_tmux_option "@dracula-gpu-expand" true)
 	if [[ "$gpu" == NVIDIA ]]; then
-    usage=$(nvidia-smi --query-gpu=utilization.gpu --format=csv,noheader,nounits | awk '{if ($1 < 10) {print "0"$1"%"} else {print $1"%"}}')
+		if $expand; then
+			usage=$(nvidia-smi --query-gpu=utilization.gpu --format=csv,noheader,nounits | awk '{if ($1 < 10) {print "0"$1"%"} else {print $1"%"}}')
+		else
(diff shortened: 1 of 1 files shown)


ethancedwards8:
I like this, just a couple of thoughts. We should give the user a choice if they want to use the average, or all the stats. Also good thinking on prefixing the 0.

Other wise, it looks good to me. If you can get that done I'll merge it!

jaywonchung:
Hi!

Thank you for the comment. I've addressed it as follows:
- The name of the new configuration knob is `@dracula-gpu-expand`, and defaults to `true`.
- Now the averaged version is also prefixed by a zero.
- Refactored the part that builds the utilization string.

ethancedwards8:
Hmm, I'm having some issues. I have a single GPU and the prefixed zero isn't there. Totally fine but I though you made it a feature.

Also the `EXPAND=$(get_tmux_option "@dracula-gpu-expand" true)`, A: Goes into the dracula.sh file, and B: need some documentation, so listed in INSTALL.md and README.md.

Another thing is whenever I try run it, it simply times out, not sure why. I can elaborate more on this if needed. And just one more minor thing you could fix while you're at it, there's some janky indentation on line 43-46, if you could fix that it'd be appreciated.

jaywonchung:
I see. Suppose the scripts aren't robust enough. I'll try to debug it, and also address other comments. Thanks!

ethancedwards8:
So what's the status for this? If it's no longer developed I'm going to close the PR :) Just wanted to ask first.

jaywonchung:
I didn't get much extra time lately. I'll close, and later submit a new one once I work on it. Thanks.

PR #73, opened 2020-09-20
Fix blocking loop when weather is not enabled

ImperadorSid:


ethancedwards8:
Ah, nice catch. Thanks.

PR #75, opened 2020-10-05
47: Enable window flags with `set -g @dracula-show-flags true`.

spoulson:
Enbolden window titles on activity or bell status.  This is more readable, IMO, than the default reverse text, especially when window flags appear in different colors.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -21,6 +21,7 @@ Customize the status bar by adding any of these lines to your .tmux.conf as desi
 * Switch from default fahrenheit to celsius: `set -g @dracula-show-fahrenheit false`
 * Enable powerline symbols: `set -g @dracula-show-powerline true`
 * Switch powerline symbols `set -g @dracula-show-left-sep ` for left and `set -g @dracula-show-right-sep ` for right symbol (can set any symbol you like as seperator)
+* Enable window flags: `set -g @dracula-show-flags true`
 * Adjust the refresh rate for the bar `set -g @dracula-refresh-rate 5` the default is 5, it can accept any number
 * Enable military time: `set -g @dracula-military-time true`
 * Disable timezone: `set -g @dracula-show-timezone false`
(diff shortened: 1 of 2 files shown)


spoulson:
This PR does the following:

* Shows window flags next to window name in the same manner as default tmux.
* Flags are colored in purple to contrast the white text and be less obtrusive to the eye.
* Powerline and non-Powerline formatting considered.
* Changed activity and bell flag formatting from "reverse" to "bold".  This eliminates poor readability in reverse mode with different text colors that turn to different background colors.  Bold will still stands out and not affect coloring.

image

ethancedwards8:
This looks great! One thing I find a bit confusing is the zoom indicator. Is htop zoomed out? Is zsh normal? I think a simple `+` or `-` would suffice. I have some mixed feelings about the bold too. Also I think the `-` on the last selected window is a bad idea.

Overall this is great. Just fix those suggestions and I'd get this merged :+1:

spoulson:
@ethancedwards8, I'm just using the standard window flags.

See: https://man.openbsd.org/OpenBSD-current/man1/tmux.1#STATUS_LINE

I could insert just the zoom indicator `Z`, but it was just as easy to insert all applicable window flags.

ethancedwards8:
I see, is there any way we could change those? I don't think they're very intuitive.

spoulson:
@ethancedwards8, I don't disagree, but that's standard tmux.

Maybe I can just keep with the scope of the issue and only put the `Z` flag for zoom.  The other flags can be sorted out in a separate issue?  I'm reluctant to jump to reengineering what's been in use for at least the last few years that I've been a tmux fan.

ethancedwards8:
That's fair, maybe @danerwilliams will have some input? I'm kind of on the fence.

danerwilliams:
@ethancedwards8 @spoulson imo we should probably just stick to standard tmux even if its not the most intuitive. Would rather conform to people's expectations.

ethancedwards8:
Okay sure thing, thanks for the PR @spoulson, I'll merge.

PR #57, opened 2020-08-06
Option to remove the time

osdamv:
A simple if to remove the time option, I do not like it because I already have a huge watch in my desktop environment

PR #81, opened 2020-10-18
Document time option toggle

Tooa:
Add missing documentation for time toggle introduced in https://github.com/dracula/tmux/pull/57


ethancedwards8:
Okay, but is there any reason there are two seperate PR's?

Tooa:
Sry, for the late response. No specific reason for separate PR's. You are right, I could have bundled them together since both changes tackle the documentation. Thanks for accepting though.

PR #83, opened 2020-10-30
change network script to works fine in all Linux flavor OS

numb95:
As in most Linux based OS, `iw` command located in `/sbin` and `/usr/sbin/iw` Normal users cannot access this tool easily, so using the `nmcli` makes this easy to show network in the status panel.

### DIFF
--- a/scripts/network.sh
+++ b/scripts/network.sh
@@ -9,7 +9,7 @@ get_ssid()
 	# Check OS
 	case $(uname -s) in
 		Linux)
-			SSID=$(iw dev | sed -nr 's/^\t\tssid (.*)/\1/p')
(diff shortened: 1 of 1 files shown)


ethancedwards8:
This puts a dependency on NetworkManager, which not all distro's use. Instead we could detect for one or the other?

ethancedwards8:
Any update on this? @numb95 I'd love to get this merged for you.

numb95:
Hi @ethancedwards8 
I just want to find another tool/solution for that. 
If I find anything good which no need for extra dependencies, I will add.
For now, I think it's better to let this issue opened till I find a better solution.

ethancedwards8:
Hi there @numb95 I've been playing around with this, I've found it takes a *really* long time to run this command, which is less than ideal. If you could figure out a faster solution I could pair my idea with it and get this fixed :)

ethancedwards8:
Closing, if anyone has an idea then lets fix it.

Issue #82, opened 2020-10-25
Enhancement: show username and hostname


wsh3776:
### Describe the feature you'd like
Is it possible to add a setting that can show the current `username` and `hostname`

### Describe risks you've considered
I think it wouldn't have much risk. 

### Describe alternatives you've considered
I can use `echo user` and `hostname` to show this, but can I show it in the bottom status bar. I often login the remote machine. So if I can see which host I am current in, it would be nice!

![image](https://user-images.githubusercontent.com/46078304/97103215-f9406180-16e5-11eb-9cdf-ec83069f3b7b.png)

I think it can be achieved by edit the `script.sh`, but I am not sure how to do it.

ethancedwards8:
Yeah we could do this, but I don't really see how useful it is. You could literally just type `whoami` and `hostname` and get those. We need to reduce feature bloat because now this project is becoming quite large and we should focus on the things that are actually needed. Like working with operating systems on than Linux and Mac, better compatibility with other plugins, color theming, etc.

Closing unless I can get a really good reason on why we need these. :)

qifei9:
I work on my laptop and a remote machine connect through ssh from my laptop at the same time. Thus I often have multiple tmux session running, some locally and some remotely. Therefore, I hope the status bar could tell me where the session is running, locally or remotely. Although I could have these info in the zsh prompt, the prompt is not always visible, for example in vim.

osbm:
What @qifei9 describes is exactly why i need this feature. Is there a way around to do this?

osbm:
Can we reopen this issue? @ethancedwards8

Theoreticallyhugo:
using `set -g @dracula-show-left-icon "#H"` one can get the host name of the machine where the session runs in the top left. what i find even more useful is `set -g @dracula-show-left-icon "#H | #S"` which tells me on which machine i am and what the name of the session is, since some of my servers have multiple sessions for different long run projects.

i think that with this in mind, we should be able to close this issue again. mby i should write about this in the docs #278

osbm:
@Theoreticallyhugo your answer is very interesting i will try that out.

I should mention, i was able to meet my need with just by doing this:
```
set -g @dracula-show-left-icon hostname
```

PR #84, opened 2020-11-13
add hostname script

drc0:


ethancedwards8:
Hm, so there have been discussions about this before, https://github.com/dracula/tmux/issues/82 . And I'm still on the fence, why do we really need it? I can just run `hostname` in a shell and I get the information this gives you. 

Right now we're focusing on reducing feature bloat, fixing any bugs, important feature requests like color customization, I just don't see a reason why we need something like this. 

If you can convince me I'll merge it, but until then I'm closing. :)

drc0:
Thanks for answering, you convinced me of the contrary . This would be only useful if there was a reliable and non hackish way to display the hostname of any host even when connecting through ssh, and this PR doesn't do that.

Il 19 novembre 2020 00:39:55 CET, Ethan Edwards <notifications@github.com> ha scritto:
>Hm, so there have been discussions about this before,
>https://github.com/dracula/tmux/issues/82 . And I'm still on the fence,
>why do we really need it? I can just run `hostname` in a shell and I
>get the information this gives you. 
>
>Right now we're focusing on reducing feature bloat, fixing any bugs,
>important feature requests like color customization, I just don't see a
>reason why we need something like this. 
>
>If you can convince me I'll merge it, but until then I'm closing. :)
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/dracula/tmux/pull/84#issuecomment-730025914

-- 
Inviato dal mio dispositivo Android con K-9 Mail. Perdonate la brevità.

PR #85, opened 2020-11-21
Added padding at the end of left icon for custom args

Memije:
When using the `dracula-show-left-icon` option with a custom character, there's missing padding to the right of the character like so

Screen Shot 2020-11-20 at 18 06 48

To fix this, the extra padding was added when assigning the `left_icon` variable


ethancedwards8:
Yeah hm, I find this kinda odd, in some terminal emulators the spacing isn't needed, but on some they are. Like I'm on Linux and using the Kitty terminal emulator and there is already a pretty big gap.

ethancedwards8:
Maybe we could make this an option? Because this is definitely annoying.

danerwilliams:
> Maybe we could make this an option? Because this is definitely annoying.

I second this

ethancedwards8:
Bumping, progress report? @Memije

Memije:
> Bumping, progress report? @Memije

@ethancedwards8 updated

ethancedwards8:
Nice, thanks!

PR #92, opened 2021-01-26
Fix: False AC charging message on MacOS

### LINKED ISSUE #91: Bug: False AC charging status
### Describe the bug

Even though charger is unplugged this plugin still reports "AC"

### To Reproduce
Steps to reproduce the behaviour:

On Macs sometimes the command used to get charging/not-charging status is this: `pmset -g batt | sed -n 2p | cut -d ';' -f 2` This can sometimes give whitespace in beginning like `" discharging"` which can cause case matching to fail in following line in battery.sh:
```bash
	case $status in
		discharging|Discharging)
```

### Expected behavior
plugin should not report AC when there's no charger attached.

### System
 - OS: Mac OS Big Sur
- Tmux Version: 3.1c

amenat:
Minor fix: removed whitespace before comparing output of battery status.

closes #91

### DIFF
--- a/scripts/battery.sh
+++ b/scripts/battery.sh
@@ -67,7 +67,7 @@ battery_status()
 		;;
 
 		Darwin)
-			status=$(pmset -g batt | sed -n 2p | cut -d ';' -f 2)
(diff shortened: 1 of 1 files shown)


ethancedwards8:
Cool, thanks.

PR #98, opened 2021-03-27
Create function to normalize cpu usage width

### LINKED ISSUE #97: Enhancement: [CPU usage fixed width]
### Describe the feature you'd like
Right now the CPU usage width is not fixed but it changes due to the number of digits:
CPU 0% (6 chars) 
vs 
CPU 11.1% (9 chars).
It seems distracting to me that all right functionalities change position consequently every 5 seconds.
It would be possible to fix the position, or to have an option to do that?

iagotito:
__Issue__: Closes #97 

__Description__:
Instead of just echo the usage percent (wich have a variable length), `get_percent()` now pass the percent to a normalizer function to set the percent string length always to 5. By doing this the CPU usage status will have a contant length, always placing its valor in the middle of the box.

__Testing__:
I commented the line that gets the percentage of the system in `cpu_info.sh` and set the percentage manually, started a new tmux session without the plugin and manually sourcered `dracula.tmux`. Then, I changed the percentage value to test all valid lengths (from 2 to "1%" to 5 to "11.1%") and waited for the script's sleep to update the status bar.
Obs.: Tested only in linux (debian 10).

### DIFF
--- a/scripts/cpu_info.sh
+++ b/scripts/cpu_info.sh
@@ -14,12 +14,24 @@ get_tmux_option() {
   fi
 }
 
+# normalize the percentage string to always have a length of 5 
+normalize_percent_len() {
+	# the max length that the percent can reach, which happens for a two digit number with a decimal house: "99.9%"
+	max_len=5
+	percent_len=${#1}
+	let diff_len=$max_len-$percent_len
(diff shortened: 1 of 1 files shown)


ethancedwards8:
LGTM

PR #107, opened 2021-06-30
Correct location of weather data file

markcornick:
This was changed in 8748783 but the main `dracula.sh` script is still looking for it in its old location, which causes tmux to hang.


ethancedwards8:
Good catch, thank you!

PR #113, opened 2021-07-26
Add support for displaying network bandwidth usage in macOS

madlep:
## Description of Changes

Major change is added support for displaying network bandwidth usage when running in macOS

While I was working on it, spotted a couple of other things to fix up
- round kB/s output to only whole numbers (constant updates to the value are annoying when it's a only a few bytes each way)
- set fixed size for displayed output width using `printf` so status bar didn't bounce around as the download/upload values changed
- spelling error the whole way through calling it "network-**bandwith**" instead of "network-**bandwidth**". Fixed that up, however, that requires renaming option names - so kept support for old options, and rewrite them to new corrected options at startup. Tried to display a message about this at startup, but `tmux display-message` doesn't do anything until a session is actually up and running.

## Testing
Have run this on macOS, and works fine, however haven't been able to test on Linux. It _should_ be ok as the only Linux specific code is kept the same.

Pic of plugin running on macOS
network_bandwidth

### DIFF
--- a/scripts/network_bandwidth.sh
+++ b/scripts/network_bandwidth.sh
@@ -2,7 +2,7 @@
 
 INTERVAL="1"  # update interval in seconds
 
-network_name=$(tmux show-option -gqv "@dracula-network-bandwith")
+network_name=$(tmux show-option -gqv "@dracula-network-bandwidth")
 
 download_bytes() {
   case $(uname -s) in
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -48,7 +48,7 @@ To enable plugins set up the `@dracula-plugins` option in you `.tmux.conf` file,
 The order that you define the plugins will be the order on the status bar left to right.
 
 ```bash
-# available plugins: battery, cpu-usage, gpu-usage, ram-usage, network, network-bandwith, weather, time
+# available plugins: battery, cpu-usage, gpu-usage, ram-usage, network, network-bandwidth, weather, time
 set -g @dracula-plugins "cpu-usage gpu-usage ram-usage"
 ```
 
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Configuration and options can be found at [draculatheme.com/tmux](https://dracul
 * Support for powerline
 * Day, date, time, timezone
 * Current location based on network with temperature and forecast icon (if available)
-* Network connection status, bandwith and SSID
+* Network connection status, bandwidth and SSID
 * Battery percentage and AC power connection status
 * Refresh rate control
 * CPU usage
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -125,6 +125,19 @@ main()
   # Status right
   tmux set-option -g status-right ""
 
+  # deprecated status right options (note spelling)
+  dep_bandwith_colors=$(get_tmux_option "@dracula-network-bandwith-colors")
+  if [ -n "$dep_bandwith_colors" ]; then
+    tmux set-option -g -o "@dracula-network-bandwidth-colors" "$dep_bandwith_colors"
+    tmux set-option -u "@dracula-network-bandwith-colors" 
+  fi
+
+  dep_bandwith=$(get_tmux_option "@dracula-network-bandwith")
+  if [ -n "$dep_bandwith" ]; then
+    tmux set-option -g -o "@dracula-network-bandwidth" "$dep_bandwith"
+    tmux set-option -u "@dracula-network-bandwith"
+  fi
+
   for plugin in "${plugins[@]}"; do
     if [ $plugin = "battery" ]; then
       IFS=' ' read -r -a colors <<< $(get_tmux_option "@dracula-battery-colors" "pink dark_gray")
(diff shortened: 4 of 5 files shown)


madlep:
Let me know if you want this squashed and/or have the separate fixes split out into different PRs

scripts/network_bandwidth.sh
```diff
@@ -0,0 +1,58 @@
...
+    sleep $INTERVAL
+
+    final_download=$(download_bytes)
+    final_upload=$(upload_bytes)
+
+    total_download_bps=$(expr $final_download - $initial_download)
+    total_upload_bps=$(expr $final_upload - $initial_upload)
+
+    total_download_kbps=$(echo "scale=0; $total_download_bps / 1024" | bc)
+    total_upload_kbps=$(echo "scale=0; $total_upload_bps / 1024" | bc)
```
o0th, line 54: `bc` in not install by default on linux, at least on my distro. What about using

```bash
    total_download_kbps=$(expr $total_download_bps / 1024)
    total_upload_kbps=$(expr $total_upload_bps / 1024)
```

scripts/network_bandwidth.sh
```diff
@@ -0,0 +1,58 @@
...
+    CYGWIN*|MINGW32*|MSYS*|MINGW*)
+      # TODO - windows compatibility
+      ;;
+  esac
+}
+
+upload_bytes() {
+  case $(uname -s) in
+    Linux)
+      cat "/sys/class/net/$network_name/statistics/rx_bytes"
```
o0th, line 24: This should be `tx_bytes`

scripts/network_bandwidth.sh
```diff
@@ -0,0 +1,58 @@
...
+      ;;
+  esac
+}
+
+main() {
+  total_download_kbps=0
+  total_upload_kbps=0
+  while true
+  do
+    echo "$(printf "%8skB/s • %8skB/s" "↓$total_download_kbps" "↑$total_upload_kbps")"
```
o0th, line 40:
```bash
    echo "$(printf "↓%8skB/s • ↑%8skB/s" "$total_download_kbps" "$total_upload_kbps")"
```
Would this be more readable? (Just personal preference)

o0th (changes requested):
I really like your changes. Could you add some line in the `INSTALL.md` regarding the config for linux users?

scripts/network_bandwidth.sh
ethancedwards8: Whats the difference between the two?

scripts/network_bandwidth.sh
ethancedwards8: Yup, we want to avoid using utilities that aren't installed by default. That includes `bc`

scripts/network_bandwidth.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
+
+INTERVAL="1"  # update interval in seconds
```
ethancedwards8, line 3: This should follow the standard that the other files follow. See the gpu_usage.sh file for a good example.

scripts/network_bandwidth.sh
```diff
@@ -0,0 +1,58 @@
...
+
+network_name=$(tmux show-option -gqv "@dracula-network-bandwidth")
```
ethancedwards8, line 5: This option needs documentation in INSTALL.md

scripts/network_bandwidth.sh
o0th: rx is for received bytes, tx is for transmitted

ethancedwards8: I see, makes sense.

scripts/network_bandwidth.sh
```diff
@@ -0,0 +1,58 @@
...
+  esac
+}
+
+upload_bytes() {
+  case $(uname -s) in
+    Linux)
+      cat "/sys/class/net/$network_name/statistics/rx_bytes"
+      ;;
+    Darwin)
+      netstat -I "$network_name" -b | tail -n 1 | awk '{print $10}'
```
ikornaselur, line 27: Would it make sense to provide the `-n` flag as well? 

I have issues currently locally with netstat hanging if I skip it, which is because (I believe, correct me if I'm wrong) it's trying to resolve addresses but we don't care about that in this case. For reference, from the man page:

```
-n    Show network addresses as numbers (normally netstat interprets addresses and attempts to 
      display them symbolically).  This option may be used with any of the display formats.
```

but having the line as `netstat -nbI "$network_name" | tail -n1 | awk '{print $10}'` works fine for me

ethancedwards8:
@madlep any updates?

madlep:
> @madlep any updates?

@ethancedwards8 hey there. Got busy with work things, and forgot about this sorry! Suggestions look good. Will work them in and update the PR.

ethancedwards8:
> > @madlep any updates?
> 
> 
> 
> @ethancedwards8 hey there. Got busy with work things, and forgot about this sorry! Suggestions look good. Will work them in and update the PR.

I completely understand, not a problem, thanks for the work!

ethancedwards8:
With the merging of https://github.com/dracula/tmux/pull/123 , this PR will have to be refactored a bit. But overall, MacOS support would be a huge plus.

PR #114, opened 2021-07-26
Fix incorrect color option being read for the time module

codeeno:
Hi there,

Currently, both the weather and time modules take their color configuration from the `@dracula-weather-colors` option, causing the time module to always look like the weather module :)


o0th (approved):
Nice catch :)

ethancedwards8:
LGTM.

PR #119, opened 2021-08-10
Fix $used_mem when page size is not 4Kb

arturts:
Page size on M1 Macs is 16k, this PR updates ram_info script to dynamically fetch the page size.

Let me know if you need me to update this PR, wasn't sure how to contribute.

### DIFF
--- a/scripts/ram_info.sh
+++ b/scripts/ram_info.sh
@@ -37,8 +37,8 @@ get_percent()
 
     Darwin)
       # percent=$(ps -A -o %mem | awk '{mem += $1} END {print mem}')
-      # Get used memory blocks with vm_stat, multiply by 4096 to get size in bytes, then convert to MiB
-      used_mem=$(vm_stat | grep ' active\|wired ' | sed 's/[^0-9]//g' | paste -sd ' ' - | awk '{printf "%d\n", ($1+$2) * 4096 / 1048576}')
+      # Get used memory blocks with vm_stat, multiply by page size to get size in bytes, then convert to MiB
+      used_mem=$(vm_stat | grep ' active\|wired ' | sed 's/[^0-9]//g' | paste -sd ' ' - | awk -v pagesize=$(pagesize) '{printf "%d\n", ($1+$2) * pagesize / 1048576}')
       total_mem=$(system_profiler SPHardwareDataType | grep "Memory:" | awk '{print $2 $3}')
       if (( $used_mem < 1024 )); then
         echo $used_mem\M\B/$total_mem


o0th (approved):
tested on linux, working fine here :+1:

ethancedwards8:
I'm testing it on Arch and the pagesize command isn't available in the base system, which I don't love. One of the goals is to avoid any external dependencies.

arturts:
Ohh bummer :/ I'll try to find another way then.

ethancedwards8:
Yeah, I'm sure the actual pagesize utility just makes a sys call, if you want to dig into the source code if it and see where that'll probably give you your answer.

ethancedwards8:
Actually, looking at it more, it only affects Darwin and doesn't change Linux functionality at all(I looked at it half asleep yesterday), so I think that this should be fine. I'll test it on my M1 MacBook Air tonight.

ethancedwards8:
Does this break backwards compatibility with Intel Macs? I don't have something to test it on. Also, on my machine `pagesize` returns 4096
```
~ ❯ pagesize
4096```

**Comment by arturts** [comment]
> Does this break backwards compatibility with Intel Macs? 

Nope, tested on my MacBook Air (2018):
```
~
❯ arch
i386

~
❯ pagesize
4096
```




>  Also, on my machine pagesize returns 4096

Are you running a terminal with Rosetta? This is what I get in a terminal running through Rosetta (M1 Mac Mini):

```
artur@artur-macmini ~ % arch
i386
artur@artur-macmini ~ % pagesize
4096
```

Same machine, non-rosetta terminal:
```
~
❯ arch
arm64

~
❯ pagesize
16384
```

ethancedwards8:
> > Does this break backwards compatibility with Intel Macs? 
> 
> 
> 
> Nope, tested on my MacBook Air (2018):
> 
> ```
> 
> ~
> 
> ❯ arch
> 
> i386
> 
> 
> 
> ~
> 
> ❯ pagesize
> 
> 4096
> 
> ```
> 
> 
> 
> 
> 
> 
> 
> 
> 
> >  Also, on my machine pagesize returns 4096
> 
> 
> 
> Are you running a terminal with Rosetta? This is what I get in a terminal running through Rosetta (M1 Mac Mini):
> 
> 
> 
> ```
> 
> artur@artur-macmini ~ % arch
> 
> i386
> 
> artur@artur-macmini ~ % pagesize
> 
> 4096
> 
> ```
> 
> 
> 
> Same machine, non-rosetta terminal:
> 
> ```
> 
> ~
> 
> ❯ arch
> 
> arm64
> 
> 
> 
> ~
> 
> ❯ pagesize
> 
> 16384
> 
> ```

Yep, that was the issue.

Issue #111, opened 2021-07-22
Create unit tests and Github actions pipeline to improve quality control / eliminate extensive manual testing


danerwilliams:
As features continue to grow, the manual testing seems to become more and more complex.

PR #120, opened 2021-08-25
Add bash script syntax check

joeperri95:
Issue
issue #111

Description of Changes
Add a simple testing pipeline to check bash syntax on scripts. This is just a basic sanity check

Testing
This can be tested with https://github.com/nektos/act


ethancedwards8:
Thanks! I'll check this out soon, this is exactly what we need. If you can think of ways to automate other things that'd be great!

PR #124, opened 2021-09-18
Added new options to use arbitrary time format.

ucchiee:
Added new options because I wanted to use my original time format.

- `dracula-arbitrary-time-format` : true/false, default to false
- `dracula-time-format` : time format, default to "%Y-%m-%d(%a) %H:%M"

Where should I write readme??


ethancedwards8:
Hmm, why are you using an if inside a case? Seems odd.

ethancedwards8:
So this is a good idea, but instead of creating two options here, could we just check if `dracula-time-format` is empty or not and go from there?

ucchiee:
> Does the code work with time zones?

Yes, you can add timezones like this:
```
set -g @dracula-time-format "%Y-%m-%d(%a) %H:%M:%S #(date +%Z)"
```

Should I change to respect the `@dracula-show-timezone` option? I think it might make this option a little bit complicated.

> Also, for backwards compatibility, let's keep the existing options at least for the time being. Maybe in the future they can be removed.

Understood.

NitroCao:
Hi, will this PR be merged in the future?

PR #132, opened 2021-10-10
Fix extra padding for time/date feature

foxtrot:
The time/date "plugin" seems to have an excess amount of padding on the end compared to the opposite side of the bar.

Here's some examples in various configurations before my patch:
Screenshot_20211010_023220
Screenshot_20211010_023239
Screenshot_20211010_023253

And here are the same configurations afterwards:
Screenshot_20211010_023120
Screenshot_20211010_023131
Screenshot_20211010_023144

Doesn't seem to cause any issues in my testing.

### DIFF
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -73,7 +73,7 @@ main()
     false)
       timezone="";;
     true)
-      timezone="#(date +%Z)";;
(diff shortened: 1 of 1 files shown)


foxtrot:
Anything blocking this? Would love to get this merged in.

PR #136, opened 2021-10-13
rewrite network_bandwidth.sh to be consistent with other script styles

xiao0yy:
1. support dracula-refresh-rate
2. default network name
3. remove infinite loop

### DIFF
--- a/scripts/network_bandwidth.sh
+++ b/scripts/network_bandwidth.sh
@@ -15,7 +15,27 @@ get_bytes()
       ;;
 
     Darwin)
-      # TODO - Darwin/Mac compatability
+      down_up_load=$(netstat -nI $1 -b | tail -n1 | awk '{print $7,$10}')
+      echo "$down_up_load"
+      ;;
+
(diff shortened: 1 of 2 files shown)


xiao0yy:
4. rewrite network_bandwidth.sh to be consistent with other script styles

scripts/network_bandwidth.sh
```diff
@@ -1,51 +1,68 @@
...
+  output_upload=$(format_bytes $total_upload_bps)
+
+  echo "↓ $output_download • ↑ $output_upload"
+}
+
+main()
+{
+  # storing the refresh rate in the variable RATE, default is 5
+  RATE=$(get_tmux_option "@dracula-refresh-rate" 5)
+  default_network_name=$(ip route | grep default | sed -e "s/^.*dev.//" -e "s/.proto.*//")
```
ethancedwards8, line 61: That will break on MacOS, `ip` is not a command. This should probably belong in its own function.

xiao0yy: I fix it and also add bandwidth read for MacOS~

Issue #34, opened 2020-05-29
Enhancement: Weather Module: Zip Code override


Sicnus:
I'm having an issue with the weather applet.  To me it seems the weather is being pulled from a city that's over 100 miles away from me? I see the code in ~/.tmux/plugins/tmux/scripts/weather.sh and I guess I could tweak that, but it'd be nice if I could simply put a zip code preference in my .tmux.conf file to override any magic IP location gathering bits:

**set -g @dracula-show-weather 31211**

Risks: Folks not in the US don't have zip codes.

Not really sure about alternatives internationally, but I'm sure someone has done this logic somewhere on plugins for different windowmanagers.

ethancedwards8:
I do believe #52 solves this issue? Try it out with the new weather plugin and we'll go from there.

abatilo:
#52 helps with international users but doesn't let you override location when IP geolocation isn't accurate. It'd still be nice if we could get this update.

PR #141, opened 2021-12-03
Enable custom location for weather plugin

adrianmihalko:
I think I am not alone who is using VPN for their work and weather plugin is showing the location from ipinfo.io, which in case of using VPN gives wrong location.

PR adds support for custom location, just enter your city. You can verify data on http://wttr.in/YOURTOWN

In tmux.conf add:

`set -g @dracula-fixed-location 'Štúrovo'`

PR also resolves issue #34

### DIFF
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -12,6 +12,7 @@ main()
   # set configuration option variables
   show_fahrenheit=$(get_tmux_option "@dracula-show-fahrenheit" true)
   show_location=$(get_tmux_option "@dracula-show-location" true)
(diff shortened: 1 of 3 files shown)


ethancedwards8:
Thank you for the PR! Will review and test, LGTM.

JannoTjarks:
Thanks for the contribution, @adrianmihalko !
I tested the code and are using it on my daily work.
It works fine for me.

ethancedwards8:
Sorry for the delay in reviewing. Thank you for your contribution!

Issue #118, opened 2021-08-10
Enhancement: Add load averages segment


martsa1:
### Describe the feature you'd like
Similar to #36, it'd be awesome to add a segment that lets us view load averages for the system.  Personally I prefer load average to CPU utilisation as I can get an indication of utilisation very recently, and over five and fifteen minute averages.

### Describe risks you've considered
There should be no risks to adding this for Linux and Mac.  I'm not sure if there's an equivalent for windows systems.

### Describe alternatives you've considered
I can live with the CPU utilisation segment already provided, but I believe Mac and Linux support for load averages should be reasonably simple:

```sh
# On Linux:
cat /proc/loadavg | cut -d ' ' -f 1,2,3

# On Mac:
sysctl -n vm.loadavg | cut -d ' ' -f 2,3,4
```

ethancedwards8:
I like it, definitely something we can implement.

pataquets:
@martsa1 Since #143 is in and working (does for me), I guess this can be closed, right?

martsa1:
I can see the changes and the PR looks good, but with the new option set in my config, I still see CPU usage, am I missing something?
```
set -g @dracula-plugins "battery cpu-usage ram-usage time"
set -g @dracula-cpu-display-load true
```
![image](https://user-images.githubusercontent.com/3308997/178141958-924c7356-df77-49b5-97ae-ee8cdf68317a.png)

martsa1:
For anyone else wanting this configuring tmux with nix, I made a custom derivation of the Dracula plugin like so (should work until a new version is released and picked up in nixpkgs):
```nix
{ pkgs, lib, ... }:

let
  dracula_head = pkgs.tmuxPlugins.mkTmuxPlugin rec {
    pluginName = "dracula";
    version = "5b282b043f760bd27b6d8c32b10d111f618b4c21";
    src = pkgs.fetchFromGitHub {
      owner = "dracula";
      repo = "tmux";
      rev = "${version}";
      sha256 = "sha256-Dhpj2NaIZO+IPsChr1uqKIR6Zv8F2QReEdU/RIVZHAc=";
    };
    meta = with lib; {
      homepage = "https://draculatheme.com/tmux";
      description = "A feature packed Dracula theme for tmux!";
      license = licenses.mit;
      platforms = platforms.unix;
      maintainers = with maintainers; [ ethancedwards8 ];
    };
  };
in
{
  programs.tmux = {
    plugins = with pkgs.tmuxPlugins; [
[...]

PR #143, opened 2021-12-22
Add load average support

adrianmihalko:
This PR adds support for load average to CPU info script.

Screenshot 2021-12-22 at 23 15 23

Also resolves request #118

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -119,11 +119,14 @@ Customize label
 set -g @dracula-cpu-usage-label "CPU"
 ```
 
-Show load average instead of percentage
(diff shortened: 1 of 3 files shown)


ethancedwards8:
How does one interpret those numbers?

adrianmihalko:
It's a quick overview of your last 1-5-15 min CPU load.

https://en.wikipedia.org/wiki/Load_(computing)

ethancedwards8:
I see, add an explanation to the docs file and I'll merge it, thanks for contributing!

ABitMoreDepth:
This looks great, thanks!

PR #146, opened 2022-01-12
utilised ram presents as a decimal value

mj-crabtree:
as title, utilised ram now features increased accuracy


ethancedwards8:
We reworked how ram is displayed and solved this issue. Thanks for the contribution though!

PR #147, opened 2022-01-26
fix: network_ping.sh need execute permission to work properly

nic-6443:


nic-6443:
@ethancedwards8 @adrianmihalko Please take a look.

tmigone:
Can confirm I had to `chmod +x ~/.tmux/plugins/tmux/scripts/network_ping.sh` for it to work. I'm on macOS

PR #149, opened 2022-02-05
Added support for system temperature on Raspberry PI

chauek:


ethancedwards8:
Is there any way to do this without using `vcgencmd`? Would love to merge this.

PR #152, opened 2022-02-11
Added kubernetes-context plugin

JannoTjarks:
This pull request is adding a plugin for kubernetes. It shows the current kubernetes status in the statusbar. It needs the tool kubectl installed. 

image

With the variable dracula-kubernetes-context-label the user can add a prefix for the statusbar entry.
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Configuration and options can be found at [draculatheme.com/tmux](https://dracul
 * When prefix is enabled smiley face turns from green to yellow
 * When charging, 'AC' is displayed
 * If forecast information is available, a ☀, ☁, ☂, or ❄ unicode character corresponding with the forecast is displayed alongside the temperature
+* Current kubernetes context
 
 ## Compatibility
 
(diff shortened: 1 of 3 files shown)


pentago:
Any chance we can get this one today? :wink wink:

JannoTjarks:
Hi @ethancedwards8  , what is the current status?

ethancedwards8:
I'm hesitant to merge this because of the use of an external tool. I realize the Git module used an external tool, but it's more mainstream and widespread.

JannoTjarks:
Hi @ethancedwards8, i understand your hesitate. But all users working with kubernetes got kubectl installed. 

As @pentago shows, there is a wish/need for a plugin like this. 
There are also existing solution like https://github.com/jonmosco/kube-tmux which will not work which dracula/tmux. The users understand, that the need kubectl as binary installed. It should be enough, to communicate it to the users through the README.

pentago:
Correct. If someone works with Kubernetes they abaolutely have kubectl installed, without it they can't interface with the API server so having kubectl is a must.

PR #156, opened 2022-03-27
Documents weather options for Tmux

rojosinalma:
It adds two options for the weather script that were not documented previously:

Set fixed location
`set -g @dracula-fixed-location "Some City"`

Hide location
` set -g @dracula-show-location false`


rojosinalma:
ping @ethancedwards8

rojosinalma:
ping @ethancedwards8

PR #159, opened 2022-05-30
Fix parameter syntax and errors

harkabeeparolus:
I fixed some syntax errors, ran ShellCheck, and cleaned up the code a little.

### DIFF
--- a/scripts/ram_info.sh
+++ b/scripts/ram_info.sh
@@ -3,7 +3,7 @@
 export LC_ALL=en_US.UTF-8
 
 current_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-source $current_dir/utils.sh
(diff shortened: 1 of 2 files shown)


ethancedwards8:
We're always looking for code improvements! Thank you. Could you rerun everything on the current code base?

harkabeeparolus:
> We're always looking for code improvements! Thank you. Could you rerun everything on the current code base?

Sure, I'm happy to help, and I'm an experienced bash developer. Would you like me to clean up the whole scripts directory while I'm at it? I noticed that [ShellCheck](https://www.shellcheck.net/) finds more errors and warnings. And in general I find that ShellCheck and [Shellharden](https://github.com/anordal/shellharden) do the right thing, making bash code more robust.

ethancedwards8:
Sure. That would be great! Thank you!

PR #163, opened 2022-07-10
Add 'custom' status plugin to get status texts from external scripts.

pataquets:
Allow custom status plugins by way of invoking arbitrary scripts (only within plugin scripts' directory).
Script output will be passed as status text (which can be also tmux's status bar placeholders, such as `#{host}`). The plugin name is `custom:`followed by the script, which must reside in the plugins dir and be executable (symlinks are OK). Otherwise an error will be shown in the plugin slot. Declaring multiple plugins works OK.
Added corresponding `@dracula-custom-plugin-colors` variable, as done in all plugins, all custom plugins have the same colour, thou, which I think it's fine (just interleave them). However, I can devise a fix, if required.

*Example*
Adding to plugins `custom:cpu_info.sh` would mimic adding the `cpu-usage` plugin. Drop your custom script or symlink in the plugins dir and adjust name as needed.

*Remaining tasks*
Documentation: Just added a notice in the readme. I guess it should go [here](https://github.com/dracula/draculatheme.com), when it's reviewed & merged.

I can also send a PR with a `command:` (or should it be `shell:`?) plugin to run custom commands (eg. `cat ~/file.txt`). However, it would require the `read -r -a` for plugins to be a `read -a` to allow spaces. I didn't send in case it's not acceptable, but it "Works on My Machine:tm:" :grin:.


ethancedwards8:
Great work. This is something we really need.

Some comments though:
Documentation that is reflected on the website is listed in the INSTALL.md file. When you add that, can you also add an example of it's usage? It's not super obvious to me. 

Also, I would be okay with allowed the script to be sourced elsewhere. I do not think it's necessary to have to put it in the scripts directory and that could make things a bit messy potentially.

ethancedwards8:
Yes, I think the responsibility ultimately falls on the user to verify that the external scripts (that are unsupported by the Dracula team) are not malicious.

PR #167, opened 2022-08-26
feat: add network-vpn plugin (only macOS)

tmigone:
Hi, added this simple check for VPN connection to my setup, figured I'd PR here in case it's of interest.

Only added support for macOS, uses `scutil` utility which I believe has been around for 10+ years.

Signed-off-by: Tomás Migone <tomasmigone@gmail.com>


ethancedwards8:
Hello, while the scutil command max exit for MacOS, it is not natively installed on most Linux systems. Could you change the script to only support MacOS while we figure out a solution for other systems?

PR #169, opened 2022-09-05
git script changes

nunojsa:
This PR adds a new option to disable untracked files from being reported as local changes. More important, the PR introduces the 
'--no-optional-locks' option. For more info, check the commit message.


nunojsa:
v2:

* Added the `--no-optional-locks'` to `getChanges()` where is also needed...

nunojsa:
@ethancedwards8 , is this theme still maintained?

PR #174, opened 2022-10-26
git: add remote info support

nunojsa:
With this new option, we get information about which remote branch we're tracking. On top of this, we'll get information about ahead/behind commits when we diverged from remote. The output format will be in the form:

'local...remote +ahead -behind',

where ahead and behind are the number of commits ahead and behind.

This functionality is controlled by a new option called '@dracula-git-show-remote-status'.

Note that for this to be properly displayed, we need to increase the size of the right status bar when the git plugin is enabled.

In order to be easier to introduce the change, getMessage() was also a bit changed in order to be easier to append the remote info.

This is how this looks like:

image

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -212,6 +212,12 @@ Hide untracked files from being displayed as local changes
 set -g @dracula-git-no-untracked-files true
 ```
 
+Show remote tracking branch together with diverge/sync state
+```bash
+# default is false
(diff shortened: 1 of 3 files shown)


nunojsa:
@danerwilliams , @ethancedwards8  any chance to look at this one?

nunojsa:
@danerwilliams , @ethancedwards8  pinging this one...

PR #176, opened 2022-10-28
Reject unrecognized plugins in config

aaronkollasch:
I like experimenting with dracula plugins, which means that my `tmux.conf` can reference plugins that aren't present in the dracula plugins directory.

In that case, dracula would repeat the preceding recognized plugin for every unrecognized plugin:
image

This PR fixes that by skipping any plugin that doesn't match:
image


ethancedwards8:
This is definitely a needed fix, however, this solution is a bit messy. Is there a more elegant fix?

aaronkollasch:
Thanks for reviewing!

PR #184, opened 2023-01-05
fix: exclude available RAM from total RAM usage

### LINKED ISSUE #183: Bug: total RAM used doesn't exclude available RAM on linux
### Describe the bug
The dracula `ram-usage` plugin on Linux doesn't exclude available RAM that can be freed when needed in total memory usage.

I use ZFS on some servers and w/o subtracting available RAM my memory usage would never be less than `total_mem/2` since ZFS allocates that much memory for disk caching if it's free.

Example output that shows the problem:
```bash
$ free -h
               total        used        free      shared  buff/cache   available
Mem:            62Gi        33Gi        18Gi       2.0Mi        10Gi        28Gi
Swap:          4.0Gi          0B       4.0Gi
```

See also: https://www.linuxatemyram.com

danielx:
The "available" column includes memory that can be made available to the system if needed but is considered as used.

See: https://www.linuxatemyram.com

This fixes #183

### DIFF
--- a/scripts/ram_info.sh
+++ b/scripts/ram_info.sh
@@ -10,17 +10,17 @@ get_percent()
   case $(uname -s) in
     Linux)
       total_mem_gb=$(free -g | awk '/^Mem/ {print $2}')
-      used_mem=$(free -g | awk '/^Mem/ {print $3}')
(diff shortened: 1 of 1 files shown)


danielx:
Closing this since i realized that the math/logic is incorrect

Issue #165, opened 2022-07-27
Bug: doesn't display ram value


babepig018:
### Describe the bug
the value of ram doesn't display on my os

### Screenshots
[![vpldII.png](https://s1.ax1x.com/2022/07/27/vpldII.png)](https://imgtu.com/i/vpldII)

### System
 - OS: Arch Linux
- Tmux Version:3.3a

yuhsienchiang:
I get a similar issue too.  CPU usage disappears and RAM usage only displays partially.

System
- OS: macOS 12.6
- Tmux: 3.3a
- Terminal emulator: Alacritty 0.10.1

<img width="1438" alt="Screen Shot 2022-09-25 at 2 33 49 am" src="https://user-images.githubusercontent.com/46809056/192108984-bca46e36-23cb-44b6-8200-baeddd7c31b7.png">

babepig018:
Solved!
check its source code
 
![2022-10-19_12-03](https://user-images.githubusercontent.com/46259718/196594994-e2698952-8b66-463f-b0fd-cf9728274f97.png)

It get data by intercepting the output of `free`

But I set the system `LANG` to Chinese,  the `free` output is chinese too

![2022-10-19_12-08](https://user-images.githubusercontent.com/46259718/196595496-902e6c66-10f2-4e2a-9485-b4a4eb04fe10.png)

so it can't get the data

the solved is change the source code in `~/.tmux/plugins/tmux/scripts/ram_info.sh`, like this
  
![2022-10-19_12-12](https://user-images.githubusercontent.com/46259718/196595980-a35c7dd9-9c80-447f-a2b8-96cf71027eea.png)

IGLOU-EU:
@babepig018 My PR #165 should solve your problem. Can you confirm? ☺️

PR #189, opened 2023-01-07
🐛 fix(#165): is no longer based on word matching

IGLOU-EU:
View issue #165 

- `free` show "Mem" every time on col 2 (hard-coded https://gitlab.com/procps-ng/procps/-/blob/master/src/free.c#L382)
- Remove useless comparison and duplicate call to `free`


ethancedwards8:
Wow, great work.

PR #190, opened 2023-01-17
continuum trigger

clanghans:
Hi, 
had a really hard time setting up continuum with dracula-theme because the status-right was overwritten. 

Thus introducing a continuum trigger option. Maybe you consider it. 

Thanks for your work.


ethancedwards8:
Sadly, support of outside themes is currently out of scope. We're looking to be able to support something like this in the future.

PR #191, opened 2023-02-10
Add OpenBSD (cpu, ram) and Fossil SCM

rbucker:
I've added some features: Add OpenBSD (cpu, ram) and Fossil SCM


ethancedwards8:
This is great, could you remove the fossil commits and file that as a separate pull request? I would love to merge the OpenBSD code right away.

PR #203, opened 2023-04-20
Fix the background agent for the weather script

### LINKED ISSUE #181: Bug: Weather unit keeps resetting to Farenheit
### Describe the bug
My tmux config has the following relevant things for `dracula/tmux`:

```tmux
set -g @plugin 'dracula/tmux'

set -g @dracula-show-powerline true
set -g @dracula-show-fahrenheit false
```

When I reload (re-source) the tmux config, the temperature uses the correct Celsius unit, however after a while (maybe when it refetches the weather? Not sure), the unit resets to F.

### To Reproduce

1. Copy my config to your `tmux.conf`
2. Wait for a while (again, not totally sure what causes it)
3. Your temperature is in Fahrenheit again

### System
 - OS: macOS 12.6.1
- Tmux Version: `tmux 3.3a`

miknikif:
## Motivation for the change
The weather daemon isn't spawned when you start Tmux.
In the condition to launch the `sleep_weather.sh` we have a check for a `tmux session` before starting while loop.
But Tmux on startup waiting for this script to finish before proceeding and creating the session.
When the session is spawned we will have weather data now because it'll be fetched by the script before the while loop is reached. After initial weather retrieval, it won't be updated anymore.

I've changed this behavior and the weather script will be triggered by the Tmux status line. It's recording the time of the last execution. If the delta between the current time and of the time of the last execution is bigger than the 1200s (default value) it won't do anything. With this change, we'll not be needed a background agent at all.

Also, `sleep_weather.sh` didn't honor theme options on the first execution (the first weather retrieval is happening without passing those options to the script). So, this PR will fix #181 

Change tested on the MacOS aarch64.


ethancedwards8:
Yes. Thanks for reminding me.

Issue #200, opened 2023-04-13
Bug: CPU plugin label can't be unset/blanked


pataquets:
### Describe the bug
Setting `@dracula-cpu-usage-label` to `""` fails to unset CPU plugin's label.
Also unsetting the variable with `set -gu @dracula-cpu-usage-label` does not work.

### To Reproduce
1. `set -g @dracula-cpu-usage-label ""`
2. Reload `tmux`
3. Label is set to default `CPU`.

### Expected behavior
Label should be blankable to save statusbar estate. I guess that the current default when blank is not preserving previous behavior.

### System
 - OS: Ubuntu Bionic
- Tmux Version: 2.6

ethancedwards8:
I think this is due to the internal implementation of get_tmux_option():

```
get_tmux_option() {
  local option="$1"
  local default_value="$2"
  local option_value="$(tmux show-option -gqv "$option")"
  if [ -z "$option_value" ]; then
    echo "$default_value"
  else
    echo "$option_value"
  fi
}
```

its checking to see if the label is empty with `-z`. When you set it to `""`, it is "empty" for bashes purposes. I'll think about a better way to do this.

ethancedwards8:
Yeah that doesn't work lol. Just did that for my own curiosity. AI sucks! ;).

Anyways, I'll think about a solution over the next few weeks. I have finals this week/next so I don't expect to make much real progress soon.

PR #204, opened 2023-04-22
fix: default value for the time format should be empty

miknikif:
Hello guys.

I've been using your theme for a while already and I've noticed that all time based options are not getting applied.
I've checked source code for the `dracula.sh` and I've noticed that option `@dracula-time-format` has a default value.

If we check time plugin source code we'll notice that if `time_format` isn't empty then it'll be used instead of the theme provided formats:
```sh
    elif [ $plugin = "time" ]; then
      IFS=' ' read -r -a colors <<< $(get_tmux_option "@dracula-time-colors" "dark_purple white")
      if [ -n "$time_format" ]; then # Time format condition right here
        script=${time_format}
      else
        if $show_day_month && $show_military ; then # military time and dd/mm
          script="%a %d/%m %R ${timezone} "
        elif $show_military; then # only military time
          script="%a %m/%d %R ${timezone} "
        elif $show_day_month; then # only dd/mm
          script="%a %d/%m %I:%M %p ${timezone} "
        else
          script="%a %m/%d %I:%M %p ${timezone} "
        fi
      fi
``` 

The problem is even if I specify an empty value for the `@dracula-time-format` option - script will use a default value. This means that this theme will always use a custom time format specified, ignoring all other options.

Please see the following screenshots:
1. What theme is showing by default right now: 
    image
2. What it should show by default:
    image

Thanks!


ethancedwards8:
Good catch. Thanks.

pataquets:
I think this applies also to #200. Looks like default values should be used with care, I guess.
/cc @miknikif @ethancedwards8

PR #205, opened 2023-04-25
fix: accept space in fixed location

tuanha168:
I add some small tweak so that fixed localtion can accept string with space like `"New York"`

### DIFF
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -180,7 +180,7 @@ main()
 
     elif [ $plugin = "weather" ]; then
       IFS=' ' read -r -a colors <<< $(get_tmux_option "@dracula-weather-colors" "orange dark_gray")
-      script="#($current_dir/weather_wrapper.sh $show_fahrenheit $show_location $fixed_location)"
+      script="#($current_dir/weather_wrapper.sh $show_fahrenheit $show_location "\"$fixed_location\"")"
 
     elif [ $plugin = "time" ]; then
       IFS=' ' read -r -a colors <<< $(get_tmux_option "@dracula-time-colors" "dark_purple white")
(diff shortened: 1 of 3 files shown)


scripts/dracula.sh
```diff
@@ -180,7 +180,7 @@ main()
 
     elif [ $plugin = "weather" ]; then
       IFS=' ' read -r -a colors <<< $(get_tmux_option "@dracula-weather-colors" "orange dark_gray")
-      script="#($current_dir/weather_wrapper.sh $show_fahrenheit $show_location $fixed_location)"
+      script="#($current_dir/weather_wrapper.sh $show_fahrenheit $show_location "\"$fixed_location\"")"
```
ethancedwards8, line 183: Couldn't this just be `"${fixed_location}"`? https://stackoverflow.com/questions/8748831/when-do-we-need-curly-braces-around-shell-variables

scripts/weather_wrapper.sh
```diff
@@ -20,7 +20,7 @@ main()
 
   if [ "$(expr ${TIME_LAST} + ${RUN_EACH})" -lt "${TIME_NOW}" ]; then
     # Run weather script here
-    $current_dir/weather.sh $fahrenheit $location $fixedlocation > "${DATAFILE}"
+    $current_dir/weather.sh $fahrenheit $location "$fixedlocation" > "${DATAFILE}"
```
ethancedwards8, line 23: Couldn't this just be `"${fixed_location}"`? https://stackoverflow.com/questions/8748831/when-do-we-need-curly-braces-around-shell-variables

scripts/dracula.sh
tuanha168: I tried `"${fixed_location}"` but no luck, still need escape quote when have space in like this in `.tmux.conf`
`set -g @dracula-fixed-location "New York"`

Just `New` got inside and the weather also wrong
image

scripts/weather_wrapper.sh
tuanha168: (duplicate comment maybe ?)

yes, it's the same with my changes

tuanha168:
@ethancedwards8 
any update ?

PR #210, opened 2023-05-11
add network public ip plugin

ljseng:
This plugin shows public IP at TMUX status line.

Color is configurable. Default foreground: cyan and background: dark_gray (inherits existing network plugin color scheme).
Label is configurable. Default empty string.  

Sample of plugin working. IP masked for privacy sake.
image


ethancedwards8:
Nice work. Some questions:
- Can you add a configuration option to configure the endpoint? 
- How often would this poll the server? This could quickly become network intensive/cpu intensive. 
- How would the program handle an absence of network connection?

Thank you for your contribution.

ethancedwards8:
Wanted to bump these requests

PR #211, opened 2023-05-12
Added eks support for kubernetes prompt

dormunis:
Makes the Kubernetes prompt more robust with the following:
* Ability to hide the ARN from the cluster name - only available for cluster whose names are ARNs
* Ability to extract account id (prefixed with account_id/<k8s-context>) - only available for clusters whose names are ARNs
* Ability to hide the user from the context (I already know who I am, this was just irrelevant and made the entire thing really bloated)

I'm well aware that the first feature (hide ARN) might be confusing, and can be controlled and renamed via kubeconfig, but I constantly move around and update my kubeconfig, between clusters of different clients, and it always overrides my existing renamed configurations.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -273,3 +273,29 @@ Set the label when there is one client, or more than one client
 set -g @dracula-clients-singular client
 set -g @dracula-clients-plural clients
 ```
+
+#### Kubernetes options
+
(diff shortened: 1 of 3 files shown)


ethancedwards8:
Thank you!

Issue #128, opened 2021-10-04
Enhancement: support different colors of active/inactive panes


ericfengchao:
### Describe the feature you'd like
support different colors of active/inactive panes. something like this

![image](https://user-images.githubusercontent.com/32474964/135806278-a243b749-d3e4-45d7-ac4d-16105ad24870.png)

ethancedwards8:
If you can find an existing implementation, please share it. I'll be happy to implement this.

o0th:
It should be easy to achieve, something like

```
set -g window-style 'fg=f8f8f2,bg=##44475a'
set -g window-active-style 'fg=#f8f8f2,bg=#282a36'
```

PR #219, opened 2023-06-16
Adding support to customize the current window colors

mrbrandao:
Those changes enables support to customize the current window foreground and background colors

+ Enabling the support to customize the current window fg/bg colors via `@dracula-current-window-colors`
+ Enabling as well support to customize the inactive window fg/bg colors via `@dracula-window-colors`

Such as:
```
set -g @dracula-current-window-colors '[background] [foregroud]' 
```

Examples:

```
# set the background of the current window status color to `dark_purple` and the foreground to `cyan`
set -g @dracula-current-window-colors 'dark_purple cyan'

# define the inactive window background to `gray` and foreground to `pink` 
set -g @dracula-window-colors 'gray pink'
```

[current-window-colors-example](https://raw.githubusercontent.com/mrbrandao/shareimg/main/tmux-dracula-theme/current-window-status-colors-example.png)

Ps: This PR might have relation with the issue #128


ethancedwards8:
How does this work if the values are not set?

Theoreticallyhugo:
i think this would be best implemented in #224 and closed

Theoreticallyhugo:
i resolved the merge conflicts, but it needs fixing. am working on it. dont merge yet :D

PR #220, opened 2023-06-17
Add tmux-ram-usage plugin

aaronkollasch:
This shows the total RAM usage of the tmux server and all child processes. ~~Requires `pstree`~~.

Example (in purple):
image

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -48,7 +48,7 @@ To enable plugins set up the `@dracula-plugins` option in you `.tmux.conf` file,
 The order that you define the plugins will be the order on the status bar left to right.
 
 ```bash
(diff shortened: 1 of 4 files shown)


ethancedwards8:
We don't generally allow PRs that require the use of an external tool (except in the case where it's immediately obvious that another binary is required like Kubernetes or Git). If you could find a way to do this without the use of pstree I would merge it.

aaronkollasch:
I added a fallback to get all child pids of the tmux server in case pstree is not installed.

In testing with a fairly small tmux session (2 panes, one neovim process running w/ LSP), the fallback function is slower than pstree on macOS, taking ~2x the time (162 ms vs. 87 ms), but it took about the same time on Ubuntu Linux (38 ms vs. 30 ms).

ethancedwards8:
That's a good compromise, thanks for your contribution!

PR #138, opened 2021-10-21
Added plugin for the window option "Synchronize panes"

JannoTjarks:
Hi guys,

i added a plugin for your tmux theme. It's called "synchronize panes" because it gives me in the rigth status bar an info whether the window option synchronize panes (:setw synchronize panes) is on or off.

I added an info to the README.md and INSTALL.md already. Maybe you have an idea for a better infotext. At the moment the label is called "Sync" in default and the value is "on" or "off". The label is customizable for the enduser in the .tmux.conf.

I hope the code meets your requirements!
grafik

PR #228, opened 2023-07-14
[fix] 'synchronize-panes' is not runnable

magnuslarsen:
Follow-up to https://github.com/dracula/tmux/pull/138

I tried it out today (thanks for the quick merge!), and I couldn't get it to work.

Turns out the script wasn't executable, and it seems like the plugin was loaded outside the plugin-loading-loop. 

So I've moved it back inside that loop, and it now works on my machine at least :-)
image


ethancedwards8:
Not sure how I missed this, thanks.

PR #231, opened 2023-08-29
fix(scripts) fix gpu usage script for NVIDIA cards

duskrunner:
For nvidia card it is displayed as "unknown" until you add this line


duskrunner:
Looks like this is also can be fixed by setting up discrete card as main for display in BIOS.

PR #233, opened 2023-10-01
check for bc program and show info if not found

dashorty:
if the bc program is missing than the bandwidth shows 0.00


ethancedwards8:
Thanks. If you want to find an alternative implementation for this functionality I would merge that as well.

PR #235, opened 2023-10-03
Add ssh-session plugin

AbelAnaya:
This plugin shows ssh connection information of active TMUX pane in the status line.

The default format is:
`user@hostname`

It can be configured to show also the connection port via `@dracula-show-ssh-session-port true`
`user@hostname:port`

When no ssh session is active it will show current user and hostname of local machine i.e: `whoami@hostname`

Color can be configured as any other dracula plugin. Default background is green and foregraund dark_gray.

Useful when user@hostname is not present in the terminal prompt or if working inside a terminal application e.g; vim, htop, etc.

Plugin appearence:
image

### DIFF
--- a/scripts/ssh_session.sh
+++ b/scripts/ssh_session.sh
@@ -39,6 +39,26 @@ get_ssh_user() {
     fi
   done
 
+  for ssh_config in `awk '
+    $1 == "Host" {
+      gsub("\\\\.", "\\\\.", $2);
+      gsub("\\\\*", ".*", $2);
+      host = $2;
+      next;
+    }
+    $1 == "User" {
+      $1 = "";
+      sub( /^[[:space:]]*/, "" );
+      printf "%s|%s\n", host, $0;
+    }' .ssh/config`; do
+    local host_regex=${ssh_config%|*}
+    local host_user=${ssh_config#*|}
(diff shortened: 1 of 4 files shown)


scripts/ssh_session.sh
```diff
@@ -0,0 +1,125 @@
...
+  local port=$(echo $1|grep -Eo '\-p\s*([0-9]+)'|sed 's/-p\s*//')
+
+  if [ -z $port ]; then
+    local port=22
+  fi
+
+  echo $port
+}
+
+get_ssh_user() {
```
ethancedwards8, line 19: Could this function be cleaned up a bit? Looks like a lot of duplicate code.

ethancedwards8 (changes requested):
Other than the function that needs to be a bit cleaner, it looks good. Thank you for your contribution!

scripts/ssh_session.sh
AbelAnaya: Function has been cleaned up and documented, thank you for pointing it out

AbelAnaya:
Thank you for your feedback, function has been cleaned up and documented.

ethancedwards8 (approved):

PR #242, opened 2023-12-06
[FIX] '~' escaping in `cwd.sh`

anosatsuk124:
I modified to remove single quotes because single quotes put between double quotes should be rendered as an actual character.

If this was the expected behavior, please close this PR 👍


ethancedwards8:
I'm not sure I follow. Can you send me screenshots of what you mean? I believe this is expected behavior.

This would change "~/Downloads" to "/home/ece/Downloads". Afaik, "~/Downloads" is expected behavior.

PR #243, opened 2023-12-15
Fix: Avoid potential delay in calc of RAM (darwin)

aftaylor2:
The system_profiler command used to determine the total memory on darwin systems performs an "Activation Lock" check. This can produce a delay in returning total memory. I began experiencing delays of around 10 seconds.

Switching from system_profiler to sysctl to calculate total memory bypasses the "Activation Lock" check / timeout scenario.

Results of "time" command using system_profiler:

> system_profiler SPHardwareDataType  0.17s user 0.23s system 3% cpu 10.165 total grep -a "Memory:"  0.00s user 0.00s system 0% cpu 10.164 total awk '{print $2 $3}'  0.00s user 0.00s system 0% cpu 10.163 total

Results of "time" command using sysctl:
> sysctl -n hw.memsize  0.00s user 0.00s system 83% cpu 0.004 total awk '{print $0/1024/1024/1024 " GB"}'  0.00s user 0.00s system 86% cpu 0.004 total

### DIFF
--- a/scripts/ram_info.sh
+++ b/scripts/ram_info.sh
@@ -19,7 +19,10 @@ get_ratio()
     Darwin)
       # Get used memory blocks with vm_stat, multiply by page size to get size in bytes, then convert to MiB
       used_mem=$(vm_stat | grep ' active\|wired ' | sed 's/[^0-9]//g' | paste -sd ' ' - | awk -v pagesize=$(pagesize) '{printf "%d\n", ($1+$2) * pagesize / 1048576}')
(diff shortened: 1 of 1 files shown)


aftaylor2:
I forgot to include the associated error message when experiencing the timeout:

> 2023-12-15 10:36:11.989 system_profiler[12578:260685] Timed out waiting for the Activation Lock Capable check

ethancedwards8:
Awesome, I much prefer this solution anyways. Thank you for the find and the contribution!

PR #246, opened 2024-01-04
Fix multi word fixed location

dwdunham:
If you set the @dracula-fixed-location to "San Diego, California" it ends up looking at "San" which is somewhere in South Korea. This occurs for any fixed-location that has a space. Something like "Portland, Oregon" can be modified to "Portland,Oregon" and still work but we can't remove the space from "San Diego". The reason that spaces in the fixed location break the backend request is that dracula.sh loses everything after the space.

Q: Why does the fixed-location end up as "San"?
A: In dracula.sh the fixed-location is not wrapped in quotes when embedded in the status-right string.

A: In weather_wrapper.sh the call to weather.sh does not wrap the fixed-location argument with quotes.

Q: When these are fixed why do we get a broken link? A: Since we are calling a web service using a url, we need to escape the spaces by changing them to %20

### DIFF
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -238,7 +238,7 @@ main()
 
     elif [ $plugin = "weather" ]; then
       IFS=' ' read -r -a colors <<< $(get_tmux_option "@dracula-weather-colors" "orange dark_gray")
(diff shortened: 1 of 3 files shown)


ethancedwards8:
Thank you for contribution! Funny bug, glad you caught it.

Issue #256, opened 2024-04-01
Bug: `network` module doesn't work on macOS 14.4 and later


pratnala:
### Describe the bug
The `airport` command has been deprecated on macOS 14.4. This leads to the `network` module not working as intended.

### To Reproduce
1. Enable network plugin in `tmux.conf` - `set -g @dracula-plugins network`
2. Reload tmux
3. See bottom right to see network name
4. Nothing is shown

### Expected behavior
Network name, if wireless, or Ethernet otherwise, must be shown instead of a blank space.

### Screenshots

<img width="162" alt="Screenshot 2024-04-01 at 10 28 54 AM" src="https://github.com/dracula/tmux/assets/1689565/5b450623-d0d4-454b-b17d-740b5aa108b9">

Network name must show in the blue box but is blank.

### System
 - OS: macOS 14.4.1
- Tmux Version: 3.2a

amitizle:
This is the alternative to `airport`;
```bash
networksetup -getairportnetwork en0 | cut -d ':' -f2 | sed 's/^[[:blank:]]*//g'
```

pratnala:
Checked in #259 thanks to @amitizle's suggestion

pratnala:
@amitizle This command now doesn't work on macOS 15. Sigh, Apple.

amitizle:
@pratnala on every major version update there's some sort of backwards compatibility issue 😵

pratnala:
@amitizle Know any other way to get network name? Not finding a simple enough way.

Theoreticallyhugo:
fixes to that are currently being discussed in #291

PR #259, opened 2024-04-08
Change how Wi-Fi network info is received on Mac

pratnala:
The `airport` command is now deprecated on macOS 14.4+. Changing it to use the `networksetup` command to get this information instead based on suggestions in #256.

### DIFF
--- a/scripts/network.sh
+++ b/scripts/network.sh
@@ -18,8 +18,8 @@ get_ssid()
       ;;
 
     Darwin)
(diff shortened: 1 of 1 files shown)


ethancedwards8:
Is this solution backwards compatible?

pratnala:
Yes, I believe it is an old BSD command so it should work on older versions of macOS (<14.4) too

ethancedwards8:
Tested it with an older device I have. Worked. Thank you!

pratnala:
This is broken in Sequoia. #291 is a possible fix.

PR #263, opened 2024-04-25
Use fixed width for network bandwidth to avoid jumps

m99coder:
Use fixed width for network bandwidth to avoid jumps.


ethancedwards8:
Much needed. Thank you!

PR #273, opened 2024-07-15
Updated playerctl script to allow scrolling text

LinkUpGames:
Updated playerctl script with a scrolling text based on the character number returned from the `playerctl_playback` variable. Right now the length is hard coded to be 25 but could be extended by updating the api for media controls.

image


ethancedwards8:
Great idea. Thank you for your contribution.

Issue #225, opened 2023-06-24
Enhancement: Use battery icon instead of label and status


Erimus-Koo:
### Describe the feature you'd like
<img width="600" alt="image" src="https://github.com/dracula/tmux/assets/17139670/83961db1-1634-44c0-ace5-7acb7c2d3c81">

There are several battery icons available in the Nerd Font, which display both the percentage and charging status.
Mine is "JetBrainsMono Nerd Font".

<img width="1228" alt="image" src="https://github.com/dracula/tmux/assets/17139670/98fe683a-29c3-4a02-8767-46c85869af7f">

Using them instead of the heart label and "AC" will save space and increase clarity.

Of course not everyone installed the nerd font, so there might be an alternative option such as "use-battery-icon".

ethancedwards8:
Good idea.

sydrawat01:
> I'm wondering is there a possibility to update the icon based on the battery percentage, charging/not charging, etc?

<img width="1083" alt="image" src="https://github.com/dracula/tmux/assets/97932746/def262d6-7d6f-422f-ad86-423bd6c4b88d">


This config works well for me!
Of course, the only caveat is the user needs to have a Nerd Font installed.

Erimus-Koo:
@sydrawat01 
I am not familiar with shell code, but this seems like a good attempt. 
Additionally, it would be nice to add color to indicate battery status - green for full charge, red for low battery - and establish some rules for when the device is charging.

ethancedwards8:
Yeah this is something to think about but hard to implement without certain fonts. I like the idea of setting a custom battery prompt.

Erimus-Koo:
I believe these glyphs should be included in all nerd fonts, but I'm not certain.
Nerd fonts look like some monospace fonts, append the icon glyphs(some standard subset), but maybe the subset has some different versions.

PR #276, opened 2024-08-18
feature/ better battery indication

Theoreticallyhugo:
in reference to #225 this pull request implements nerdfont icons as battery indicators. it shows how much and whether the pc is charging. additionally the pull request is adding a few extra cases i stumbled upon in macos.

please let me know if my change to the readme is sensible or should be altered to not break the pattern.

special thanks to @sydrawat01 whose code i partially copied to replace my old solution with.


ethancedwards8:
I love this one, thanks!

PR #279, opened 2024-08-19
feature/ MacOS show vpn name and tailscale exit node if active

Theoreticallyhugo:
on macos and linux the vpn option now is able to show the name of the vpn if in verbose mode or `VPN` if verbose is off.
if a tailscale exit-node is used, it shows its name in verbose, or `Tailscale` if verbose is off.

scutil did some funky stuff for me, so i had to filter out the asterisks 😅

new options for the network-vpn plugin
`@dracula-network-vpn-label` 
this option is what is displayed when verbose is false.
if verbose is true, this label is the prefix to the name of the vpn/ exitnode.
`@dracula-network-vpn-verbose`
whether to display the vpn/ exitnode name

### DIFF
--- a/scripts/network_vpn.sh
+++ b/scripts/network_vpn.sh
@@ -17,14 +17,47 @@ vpn_function() {
       echo "NO VPN"
     fi
   ;;
-  
-  Darwin)
-    vpn=$(scutil --nc list | grep Connected)
 
-    if [ -z $vpn ]; then
+  Darwin)
+
+    verbose=$(get_tmux_option "@dracula-network-vpn-verbose" false)
+
+    vpn="$(scutil --nc list | sed "s/\*//g" | grep Connected)"
+
+    is_not_tailscale=$(echo "$vpn" | grep -v Tailscale)
+
+    which -s tailscale > /dev/null
+    tailscale_installed=$?
+
+    if [ -z "$vpn" ]; then
       echo ""
+
+    elif [ $tailscale_installed ] && [ -z "$is_not_tailscale" ]; then
+      # if tailscale is installed and no other vpn is connected. this is because tailscale will
+      # always show as connected for some reason.
+      #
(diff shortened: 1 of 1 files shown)


scripts/network_vpn.sh
```diff
@@ -5,26 +5,90 @@ export LC_ALL=en_US.UTF-8
...
+    which -s tailscale > /dev/null
+    tailscale_installed=$?
+
     if [[ $vpn =~ ^([0-9]{1,3}\.){3}[0-9]{1,3}$ ]]; then
       echo $vpn
+    elif [ $tailscale_installed ]; then
+      # if tailscale is installed
+      #
+      # https://www.reddit.com/r/Tailscale/comments/18dirro/is_there_a_way_i_can_tell_which_exit_node_i_am/
+      node=$(tailscale status --peers --json | jq '.ExitNodeStatus')
```
ethancedwards8, line 28: I'm hesitant to rely on jq as a dependency.

Theoreticallyhugo: understandable. im planning to rework this entirely if i end up making it "production ready" and i will keep that in mind then :+1:

ethancedwards8: I was poking around with this and I think you could just replace this with grep since you're not actually interpreting the json and just checking if it exists. I tested it with tailscale (I am now using it) and doesn't even report `ExitNodeStatus` unless you are connected to an exit node @Theoreticallyhugo

Theoreticallyhugo:
so far this is tested on macos only, and i'll need to look into vpns some more before this is ready to merge

ethancedwards8:
I'll test on Linux tonight. I'll also test on MacOS. For now, tailscale support is fine. We can support others as they come.

ethancedwards8 (approved):
Works for me on Linux on Darwin

PR #280, opened 2024-08-22
better battery indication for charging from AC

mamantoha:
Follow-up to https://github.com/dracula/tmux/pull/276.

However, the icons for charging from AC have been replaced with a single-symbol icon instead of two symbols for better visibility.

image

### DIFF
--- a/scripts/battery.sh
+++ b/scripts/battery.sh
@@ -112,17 +112,17 @@ battery_status()
     charging|Charging)
       # charging from AC
       declare -A battery_labels=(
-        [0]="󰂎"
-        [10]="󰁺"
-        [20]="󰁻"
-        [30]="󰁼"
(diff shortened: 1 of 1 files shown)


mamantoha:
image

Theoreticallyhugo:
awesome! any clue why linux has no percent symbol? seems like that needs fixing (not trying to say that it was you).
if i could approve i would :D

mamantoha:
The master branch is also missing the "%" symbol. I will prepare a fix in a separate PR.

image

Theoreticallyhugo:
cool :D

mamantoha:
https://github.com/dracula/tmux/pull/281

ethancedwards8:
Thank you so much!

PR #283, opened 2024-08-28
fixed weather app showing "Location Unavailable"

ikubetoomuzik:
My work environment blocks outgoing ping so the weather applet has never worked. 
I've changed the weather script to use Linux's TCP stack to test connection to port 443 of ipinfo.io instead of pings.

I also added a space for formatting the Unicode symbol and the temperature as on my system they were overlapping, I am happy to revert that change if requested.


ethancedwards8:
Okay, I like this. It makes sense. I never considered a situation like that.

Issue #278, opened 2024-08-19
Enhancement: documentation


Theoreticallyhugo:
I would like to write documenation for all the optional flags that can be set for this plugin. i'm thinking of a docs directory with a markdown file for each individual feature, though possibly grouping things like the gpu features. 

before putting in the effort to write all this however, i would like to know if that is something that would be merged if done well.
please let me know and i'll get to work :D

ethancedwards8:
Yes. I think it is much needed.

Theoreticallyhugo:
aight, on it :D

Theoreticallyhugo:
i am still working on the documentation, which you can follow here:
https://github.com/Theoreticallyhugo/dracula-tmux/tree/feature/docs/docs

a bunch of things are not done, or not included yet.

additionally, im planning on picking up #224. as a part of that i'd like to add colour customisation to more places, as well as putting the variables in a separate file, allowing for easily using fully custom colours. i will be adding these features over time, whilst documenting them too.

esfxra:
This website used to have helpful details on how to configure the theme and the plugins https://draculatheme.com/tmux

But it is no longer available

Theoreticallyhugo:
dependencies may be an interesting project for the docs, but ill close this issue for now

PR #284, opened 2024-08-28
feature/hide ssh-session when not conntected

Theoreticallyhugo:
adding an option that allows to hide the ssh host when not connected via ssh.
so far i found it confusing when my laptops host was displayed in the ssh_session flag. this PR adds an option (that per default is deactivated) to hide the flag when not connected via ssh.

documentation follows in #278


ethancedwards8:
Thank you! This was definitely needed.

PR #285, opened 2024-08-30
Add new kubernetes option to just show the context name

josepedropereira:
When using the kubernetes plugin the existing options create very extensive labels and hard it's to differenciate between them if you have several environments in the same cloud account.
Therefor the kubernetes config context name might be the best option to differentiate between them.


ethancedwards8:
Makes sense. Thank you for the contribution!

josepedropereira:
> Makes sense. Thank you for the contribution!

Thank you, the codebase is quite clean and a pleasure to work on!

PR #286, opened 2024-09-01
Feature/better gpu

Theoreticallyhugo:
- adding support for apple M chip gpus (usage and power, but not vram) 
current solution sadly requires sudo privileges. if i can find something better, i will make another pull-request
- changing to nvidia gpu detection by checking for presence of nvidia-smi tool. 
if nvidia-smi is not installed, it doesnt matter whether an NVIDIA gpu is installed, since we couldnt read the data anyways.
if it is installed, theres most probably also a gpu, because why would you install nvidia drivers otherwise?
- changing from dracula-ignore-lspci to dracula-force-gpu option.
by default its set to false, and doesnt do anything, but you can set it to "NVIDIA", making it as if NIVDIA was detected.
hopefully this will not need to be used for nvidia anymore with the switch to `type -a nvidia-smi` for detection.
- adding percentage option for vram and power
- adding option for finer grained info on vram usage

since i neither own an intel, nor amd gpu, i cannot test these changes for the respective brands. however im planning on teaming up soon to tackle that too.

roadmap:
- intel/ amd support
- multiple gpu support 
- adding all this info to the documentation

### DIFF
--- a/scripts/gpu_power.sh
+++ b/scripts/gpu_power.sh
@@ -49,7 +49,7 @@ main()
 {
   # storing the refresh rate in the variable RATE, default is 5
   RATE=$(get_tmux_option "@dracula-refresh-rate" 5)
-  gpu_label=$(get_tmux_option "@dracula-gpu-usage-label" "GPU")
+  gpu_label=$(get_tmux_option "@dracula-gpu-power-label" "GPU")
   gpu_usage=$(get_gpu)
   echo "$gpu_label $gpu_usage"
   sleep $RATE
--- a/scripts/gpu_power.sh
+++ b/scripts/gpu_power.sh
@@ -20,7 +20,10 @@ get_platform()
       ;;
 
     Darwin)
-      # TODO - Darwin/Mac compatability
+      # WARNING: for this to work the powermetrics command needs to be run without password
+      #   add this to the sudoers file, replacing the username and omitting the quotes.
+      #   be mindful of the tabs: "username		ALL = (root) NOPASSWD: /usr/bin/powermetrics"
+      echo "apple"
       ;;
 
     CYGWIN*|MINGW32*|MSYS*|MINGW*)
@@ -28,13 +31,14 @@ get_platform()
       ;;
   esac
 }
-
 get_gpu()
 {
   gpu=$(get_platform)
   if [[ "$gpu" == NVIDIA ]]; then
     usage=$(nvidia-smi --query-gpu=power.draw,power.limit --format=csv,noheader,nounits | awk '{ draw += $0; max +=$2 } END { printf("%dW/%dW\n", draw, max) }')
 
+  elif [[ "$gpu" == apple ]]; then
+    usage="$(sudo powermetrics --samplers gpu_power -i500 -n 1 | grep 'GPU Power' | sed 's/GPU Power: \(.*\) \(.*\)/\1\2/g')"
   else
     usage='unknown'
   fi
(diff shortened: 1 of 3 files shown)


ethancedwards8:
Thank you for your extensive contribution. As this is such a large pull request, I would feel better if there was documentation for the new options, specifically for the apple GPU powermetrics option. I find it unlikely that most people will look in the docs to see that they must add the command into their sudoers file. 

> if it is installed, theres most probably also a gpu, because why would you install nvidia drivers otherwise?
Some Linux distributions may come with them preinstalled.

Theoreticallyhugo:
im definitely adding all the options to the documentation im currently working on. 
additionally i could add the sudoers note into the readme, so that people who add the gpu option will have to see it.
regarding the nvidia-smi option i have an idea.
what if checking for nvidia-smi is a fallback option if the lspci command doesnt yield any result?

ethancedwards8:
> im definitely adding all the options to the documentation im currently working on. additionally i could add the sudoers note into the readme, so that people who add the gpu option will have to see it. 

Please do.
> regarding the nvidia-smi option i have an idea. what if checking for nvidia-smi is a fallback option if the lspci command doesnt yield any result?

I am good with that.

scripts/gpu_power.sh
```diff
@@ -45,7 +57,7 @@ main()
 {
   # storing the refresh rate in the variable RATE, default is 5
   RATE=$(get_tmux_option "@dracula-refresh-rate" 5)
-  gpu_label=$(get_tmux_option "@dracula-gpu-usage-label" "GPU")
+  gpu_label=$(get_tmux_option "@dracula-gpu-power-label" "GPU")
```
ethancedwards8, line 60: This would break any existing configs. What is your rational for changing this?

scripts/gpu_ram_info.sh
```diff
@@ -32,19 +35,29 @@ get_platform()
...
-  normalize_percent_len $usage
+  echo $usage
 }
 
 main()
 {
   # storing the refresh rate in the variable RATE, default is 5
   RATE=$(get_tmux_option "@dracula-refresh-rate" 5)
-  gpu_label=$(get_tmux_option "@dracula-gpu-usage-label" "VRAM")
+  gpu_label=$(get_tmux_option "@dracula-gpu-vram-label" "VRAM")
```
ethancedwards8, line 60: Maybe VRAM should be its own config option and usage stay as it is.

Theoreticallyhugo:
currently we have three different widgets using the @dracula-gpu-usage-label option: usage, power, and vram. that doesnt make a lot of sense to me, especially since for vram it defaults to "VRAM" and power and usage default to "GPU". this differentiation of labels is lost as soon as the option is set.
i think that each of the widgets should have its individual option, which is why i changed the option for two of them.

Theoreticallyhugo:
nvidia-smi detection is now used as fallback option.
tested on manjaro with no lspci issues
tested on ubuntu with no lspci installed. uses nvidia-smi as fallback
tested on macos with an M1

ethancedwards8:
Thank you so much! I appreciate all the work you have done recently.

Theoreticallyhugo:
and theres more work to come!
in private ive been maintaining a fork for more than a year now, but i havent really felt confident to contribute to a public project yet. now that i do, theres a backlog of features id like to contribute, especially since id like to switch to using the original repo instead of my fork in the future.

ethancedwards8:
Well everything you've submitted so far has been great! Keep doing what you're doing.

PR #289, opened 2024-09-10
weather.sh: added timeouts to curl requests

ikubetoomuzik:
wttr.in occasionally has issues where requests hang, this can cause issues in the status bar.

changed the script so it exports defined functions and calls them in a sub-shell. This allows for use of the timeout command on bash functions.

### DIFF
--- a/scripts/weather.sh
+++ b/scripts/weather.sh
@@ -35,6 +35,7 @@ fetch_weather_information()
 #get weather display
 display_weather()
 {
+  fahrenheit=$1
   if $fahrenheit; then
     display_weather='&u' # for USA system
   else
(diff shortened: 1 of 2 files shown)


ethancedwards8:
Brilliant. Thanks.

mamantoha:
Hi @ethancedwards8. Please revert this PR as it completely breaks the weather display.

scripts/weather.sh
```diff
@@ -66,13 +67,24 @@ forecast_unicode()
...
+export -f display_location
+export -f forecast_unicode
+export -f fetch_weather_information
+
 main()
 {
   # process should be cancelled when session is killed
   if timeout 1 bash -c "</dev/tcp/ipinfo.io/443" && timeout 1 bash -c "</dev/tcp/wttr.in/443"; then
-    echo "$(display_weather)$(display_location)"
+    if ! weather=$(timeout 3 bash -c "display_weather $fahrenheit"); then
```
mamantoha, line 79: `$(display_location)` is missing here.
Why `$fahrenheit` only?

PR #291, opened 2024-09-17
Patch to fix #290

### LINKED ISSUE #290: Bug: [You are not associated with an AirPort network.]
### Describe the bug
Probably not a bug with dracula.

In macOS 15 Sequoia, `networksetup -getairportnetwork [device]` no longer replies with the actual network...
It just says: `You are not associated with an AirPort network.`

### System
 - OS: macOS 15 Sequoia
- Tmux Version: doesn't matter...

jby:
Since Apple apparently has changed it's functionality

### DIFF
--- a/scripts/network.sh
+++ b/scripts/network.sh
@@ -22,9 +22,9 @@ get_ssid()
       ;;
 
     Darwin)
-      if networksetup -getairportnetwork en0 | cut -d ':' -f 2 | sed 's/^[[:blank:]]*//g' &> /dev/null; then
+      if system_profiler SPAirPortDataType | awk '/Current Network Information:/ { getline; print substr($0, 13, (length($0) - 13)); exit }'; then
         wifi_label=$(get_tmux_option "@dracula-network-wifi-label" "")
-        echo "$wifi_label$(networksetup -getairportnetwork en0 | cut -d ':' -f 2)" | sed 's/^[[:blank:]]*//g'
+        echo "$wifi_label$(system_profiler SPAirPortDataType | awk '/Current Network Information:/ { getline; print substr($0, 13, (length($0) - 13)); exit }')"
       else
         echo "$(get_tmux_option "@dracula-network-ethernet-label" "Ethernet")"
       fi
--- a/scripts/network.sh
+++ b/scripts/network.sh
@@ -22,9 +22,9 @@ get_ssid()
       ;;
 
     Darwin)
-      if system_profiler SPAirPortDataType | awk '/Current Network Information:/ { getline; print substr($0, 13, (length($0) - 13)); exit }'; then
+      if ipconfig getsummary en0 | awk -F ' SSID : ' '/ SSID : / {print $2}'; then
         wifi_label=$(get_tmux_option "@dracula-network-wifi-label" "")
-        echo "$wifi_label$(system_profiler SPAirPortDataType | awk '/Current Network Information:/ { getline; print substr($0, 13, (length($0) - 13)); exit }')"
+        echo "$wifi_label$(ipconfig getsummary en0 | awk -F ' SSID : ' '/ SSID : / {print $2}')"
       else
         echo "$(get_tmux_option "@dracula-network-ethernet-label" "Ethernet")"
       fi


ethancedwards8:
I verified this on MacOS Ventura. I am several versions behind lol. One concern I have is that the command hangs. Can other users test this and provide feedback?

ethancedwards8:
It may just be my machine. I will test it on another one I have at home.

Theoreticallyhugo:
im running sonoma 14.6.1 on an m1 air 8gb and the command takes about 3.7s to complete. 
however both commands seem to run fine on my machine, as soon as i fix the error with the misplaced )" in line 27.
Screenshot 2024-09-21 at 22 53 20

jby:
@Theoreticallyhugo - on macOS 14 Sonoma the `networksetup -getairportnetwork` *does* work, but not on macOS 15 Sequoia, it just answers:
```
time networksetup -getairportnetwork en0
You are not associated with an AirPort network.
user=0.01s system=0.01s cpu=45% total=0.036
```
Hence the need for a different solution in Sequoia

Theoreticallyhugo:
i see the issue, however im curious. does system profiler also have the significant delay on sequoia? 
i would think that making an OS check and using the faster command based on that may be an idea.

jby:
> i see the issue, however im curious. does system profiler also have the significant delay on sequoia? i would think that making an OS check and using the faster command based on that may be an idea.

A _complete_ `system_profiler` run on Sequoia takes 12,80 seconds
If run for _only_ SPAirPortDataType `system_profiler SPAirPortDataType` it takes 0,09 seconds

dot-1q:
If I may chime in, `system_profiler SPAirPortDataType` takes roughly 3 seconds on my machine (MBA M2) running Sequoia, which is just too much.
I wonder if we can achieve the same behavior via another command.
From my *very limited* testing, `networksetup -listpreferredwirelessnetworks en0` seems to output the connected Wifi's SSID at the top of the list. It would then be a matter of formatting the output. If others can also replicate this behavior, it could be worth pursuing this route.

Theoreticallyhugo:
@dot-1q s ipconfig based command works very quickly on Sonoma. if that also works on Sequoia, i'd say that it's the command to go with.

litszwaiboris:
> If I may chime in, `system_profiler SPAirPortDataType` takes roughly 3 seconds on my machine (MBA M2) running Sequoia, which is just too much. I wonder if we can achieve the same behavior via another command. From my _very limited_ testing, `networksetup -listpreferredwirelessnetworks en0` seems to output the connected Wifi's SSID at the top of the list. It would then be a matter of formatting the output. If others can also replicate this behavior, it could be worth pursuing this route.
> 
> EDIT: Found that `ipconfig getsummary en0 | awk -F ' SSID : ' '/ SSID : / {print $2}'` also works. Credit: [FelixKratz/SketchyBar#616 (comment)](https://github.com/FelixKratz/SketchyBar/issues/616#issuecomment-2370891970)

yay its my issue

jby:
The `system_profiler` way (the one in my patch) works for both macOS14 and macOS15 whereas the `ipconfig getsummary en0` only works on macOS15...

Theoreticallyhugo:
@jby I'm running Sonoma 14.6.1 and `ipconfig getsummary en0 | awk -F ' SSID : ' '/ SSID : / {print $2}'` gets the right output very quickly

dot-1q:
> The `system_profiler` way (the one in my patch) works for both macOS14 and macOS15 whereas the `ipconfig getsummary en0` only works on macOS15...

The `ipconfig` command comes with macOS by default if im not mistaken, so it works with both macOS14 & 15

jby:
> @jby I'm running Sonoma 14.6.1 and `ipconfig getsummary en0 | awk -F ' SSID : ' '/ SSID : / {print $2}'` gets the right output very quickly

Ok

litszwaiboris:
although this can be run directly on terminal with quick succession, for me, I can't run this within a script since it would return as empty

ethancedwards8:
> @dot-1q s ipconfig based command works very quickly on Sonoma. if that also works on Sequoia, i'd say that it's the command to go with.

I agree. It also works quickly for me too. @jby , if you would like to edit the PR to use that command, I'll get it merged ASAP. Thank you everyone!

jby:
@ethancedwards8 - updated!

PR #294, opened 2024-09-25
feature/customizable color codes

omeyenburg:
This should introduce more customizability regarding the colors.
Right now the colors (e.g. "white", "gray") have hard-coded hex color codes in `dracula.sh`. Though some users (like me :p) might want to change individual colors to match their preferences or are only interested in the status bar and not the color scheme itself.

With these changes, you can directly set the hex color code in the tmux.conf file like this:
`set -g @dracula-color-[color name] "#[hex code]"`

So changing the hex code of gray to pitch black would just be:
`set -g @dracula-color-gray "#000000"`

### DIFF
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -37,17 +37,17 @@ main()
   show_empty_plugins=$(get_tmux_option "@dracula-show-empty-plugins" true)
 
   # Dracula Color Pallette
-  white='#f8f8f2'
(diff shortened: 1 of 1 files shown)


ethancedwards8:
I'm on the fence about this. I like the idea, but this _is_ a Dracula themed Tmux plugin, which has certain associated colors.

I think I'll go ahead and merge it anyways.

PR #224, opened 2023-06-24
Improve customization

Erimus-Koo:
I took the liberty of rearranging the code and made a lot of changes at once. I hope these changes are useful and not too difficult to understand.

Because I work as an artist in a gaming company, the changes mainly focus on spacing, color, icons, etc.

image

## Rearrange

The previous code may have gone through multiple additions, so after I made changes, the new code I added looked even more messy. So in the end, I adjusted the order of the code and its naming consistency. However, I did not adjust the name of tmux's options to ensure backward compatibility.

The new code follows the following rules.

- Color Palette
- Get Tmux Option Variables
    - General
    - For Left Icon
    - For Window
    - For Plugins
- Left Icon Area
- Window Area
- Right Plugins Area

## Left Icon Area

- Add padding & margin
- Add customized color for prefix status (Previous is )
- Remove too many if, and standardize the logic for enabling or disabling powerline.
- status_bg -> variable

Here is my configuration, and **all default values that have not been changed**.
Left-Icon

### Handling of show_powerline

In the previous version, there were many `if` statements used to handle `show_powerline`.
However, I believe it is simply a marker. If `show_powerline` is not enabled, changing the marker to blank (`""`) would suffice. Although there are still several color operations of the blank that need to be handled, it is better than duplicating everything.

## Window Area

- Customize color
- Customize separator (can invert left sep color)
- Customize padding & margin
- Remove the color of flags (follow the fg)
- The window tab will maintain a consistent width regardless of focus.
  The tab will not suddenly change while you switch between windows.
  (Flags will change the width)

Window

### Disable dracula's window settings

If you turn `@dracula-window-disabled` to true, all the window settings in `dracula.sh` won't be rendered, and `.tmux.conf` works. You can do some complex things as you wish.

image

- Index: Window name > folder of the current pane
- You can also display the number of panes, current program, etc.

## Plugins Area

- Customize each side padding (rightmost separately)
- Fix the padding of the timezone string.
  If the timezone existed, add one space before the timezone. If not, there's nothing.
  (Prev version always adds an extra space)

Plugins

## P.S.

In my daily work, I primarily write Python and front-end projects, and I rarely write Shell scripts. So, I am concerned about making errors and would greatly appreciate a review of my work.

PR #296, opened 2024-09-26
Feature/theming

Theoreticallyhugo:
the idea here is to allow fully custom colours, both in the hex-code, as well as the name and number of available colours, whilst keeping changes minimal.

i created this in an attempt to slowly but surely recover and go towards the goal of #224.

changes are documented on the documentation branch im working on, including some info to use other themes with dracula

what changed:
the colour variables now live in a separate file, which is sourced per default. a new colours.sh file can be created where the user wants ( for example `~/.config/tmux/colours.sh`) and then sourced with the option `set -g @dracula-colors "$current_dir/colors.sh"`

so, for default users theres very little changes under the hood and none that they could observe. 
however this little changes opens up many possibilities for customisation.


ethancedwards8:
I'm going to close this in favor of #224 

I would appreciate any feedback you had on that PR.

Theoreticallyhugo:
I'll look into it. however I'm somewhat saddened with the lack of discussion of pros and cons between #294 and #296 😅

Theoreticallyhugo:
but thats an issue i kept in mind and actually the reason i created this solution in the first place. this would allow a user to create a file with whatever variables in any location on their machine. so, if a user had their tmux.conf under vcs, they could put the colours.sh file in the same dir and point the tmux option to that file. 
basically the idea is to have one tmux option to set all colours at once, without having to touch the programs code at all. this could also be achieved by a long string appended to the tmux option, but would be quite ugly to read. hence the external file solution.

ethancedwards8:
That is a very good point. I did not consider having the file outside of the plugin directory. I just assumed that it would be the one to be edited. 

I think I would be more open to this idea if instead of having a file in the directory, which can be misleading, we leave the default colors in dracula.sh but then overwrite them if the script detects the presence of a configuration file. That way, users are not tempted to edit the colors.sh file in the plugin repo, causing issues with updating. 

We would also need to provide an example configuration (essentially what you currently have in colors.sh right now) in INSTALLATION.md. I would not merge this without it.

Theoreticallyhugo:
i like the idea with keeping the default colours and just overriding them. i realised earlier today that not doing it that way may lead to weird issues and crashes when a user forgets assigning a variable that is used somewhere 😅

im gonna look into solutions, where the info could be placed inside the tmux.conf.

besides ive been working on configuration explanations/ installation instructions/ templates for themes in my documentation branch. i know a few people who would love to use this plugin if they could use it with guvbox or catppuccin and so im working on default configs that can be copied.

i can make sure to put instructions into installation.md for now, but im also assuming that it may move when the bigger documentation is getting ready to get merged.

i'm very happy that this solution may get merged. i've been hoping for this day for the past two years 😅

Theoreticallyhugo:
converting this to a draft, for this is currently work in progress.
i've added info to the readme and install.md, pointing to the instructions in the docs dir.
i'm currently testing the catppuccin values, and have asked a friend to check gruvbox, as he knows that better than me.
both of those plug and play configs are configured without #224 ofc. we could consider implementing #224 in this pr, or wait with this pr till the #224 successor is merged, in order to finalise documentation on color theming. for that id like to hear your thoughts @ethancedwards8.

im still looking for a solution that would allow us to configure everything in tmux.conf.

INSTALL.md
```diff
@@ -126,6 +126,14 @@ set -g @dracula-powerline-bg-transparent true
 set -g @dracula-inverse-divider 
 ```
 
+#### [color theming](https://github.com/dracula/tmux/blob/master/docs/color_theming/README.md)
```
ethancedwards8, line 129: This needs to be a relative link instead of an absolute link: https://stackoverflow.com/questions/7653483/github-relative-link-in-markdown-file

INSTALL.md
```diff
@@ -126,6 +126,14 @@ set -g @dracula-powerline-bg-transparent true
...
+
+Each individual widget's foreground and background color can be overridden.
+Additionally, the variables used for storing color values can be overridden and extended.
+This allows for the use of custom themes like catppuccin or gruvbox.
+
+For everything regarding colors, please refer to [the color theming directory](https://github.com/dracula/tmux/blob/master/docs/color_theming/README.md).
```
ethancedwards8, line 135: See above

scripts/colors.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
ethancedwards8, line 1: Remove this file.

scripts/dracula.sh
```diff
@@ -39,17 +39,19 @@ main()
...
+  light_purple="#bd93f9"
+  dark_purple="#6272a4"
+  cyan="#8be9fd"
+  green="#50fa7b"
+  orange="#ffb86c"
+  red="#ff5555"
+  pink="#ff79c6"
+  yellow="#f1fa8c"
+
+  source "$(get_tmux_option "@dracula-colors" "$current_dir/colors.sh")"
```
ethancedwards8, line 54: Remove setting the default color file to the one in the scripts directory. There should not be a default as per the reasons we discussed yesterday.

README.md
```diff
@@ -16,6 +16,7 @@ Configuration and options can be found at [draculatheme.com/tmux](https://dracul
 
 - Support for powerline
 - Day, date, time, timezone
+- [Fully custom color theming](https://github.com/dracula/tmux/blob/master/docs/color_theming/README.md)
```
ethancedwards8, line 19: See above

docs/color_theming/README.md
```diff
@@ -0,0 +1,58 @@
+# overriding widget colors
```
ethancedwards8, line 1: This is A LOT. Having all of these changes will significantly bloat the dracula.sh file.

ethancedwards8:
> converting this to a draft, for this is currently work in progress. i've added info to the readme and install.md, pointing to the instructions in the docs dir. i'm currently testing the catppuccin values, and have asked a friend to check gruvbox, as he knows that better than me. both of those plug and play configs are configured without #224 ofc. we could consider implementing #224 in this pr, or wait with this pr till the #224 successor is merged, in order to finalise documentation on color theming. for that id like to hear your thoughts @ethancedwards8.
> 
> im still looking for a solution that would allow us to configure everything in tmux.conf.

Thanks for adding documentation. 

Generally, it is advisable to do one thing at a time in FOSS projects. It keeps Git history clean, and makes it easier on maintainers and users to digest what is happening. For that reason, I think we should keep #224 and this PR strictly separate. 

I'm not sure that the individual theming of each widget (git, kubernetes, terraform, etc) is necessary? Doesn't just setting the global colors change the widget colors as well?

Theoreticallyhugo:
i saw that you commented and will look into fixing later.

the individual theming of each widget is a feature that is not added by this pr and which i still consider important. there are cases in which multiple widgets use the same color by default, which is kinda ugly/ weird when they're next to each other. take for example all network widgets, which come in the same colors. global theming would change the color of each widget in the same way, which wouldn't solve the issue of indistinguishability. 
besides, it is a very nice feature for a color theme and we do already have it implemented.

docs/color_theming/README.md
Theoreticallyhugo: since these options are somewhat scattered in the dracula.sh file, which has been that way for quite some time, i thought it would be sensible to not change any of that, but write easier to read documentation about it. considering that its been a somewhat undocumented feature for a long time, i think that while there probably are quite a few people using these options (including myself), it should've been easier to find them.

Theoreticallyhugo:
@ethancedwards8 i fixed the comments you had and changed the code to get rid of the external file. 
i'm still working on documentation for the plug n play themes (only catppuccin and gruvbox so far), but the theming option itself should be fully functional now. will do some testing

docs/color_theming/README.md
ethancedwards8: I agree.

ethancedwards8:
Everything here looks really good. I've thought about everything for a few days, and I can't really say that there is anything I would change. I am open to hearing feedback from other contributors.

Theoreticallyhugo:
awesome! ill add more prebuilt theme stuff when we have merged a pr that allows for more customisation :D

PR #297, opened 2024-10-04
Fix: get status and capacity with multiple power_supplies

NeumoNeumo:
It is possible that $BAT contains multiple power_supplies like
```bash
/sys/class/power_supply/AC
/sys/class/power_supply/BAT0
/sys/class/power_supply/ucsi-source-psy-USBC000:001
```
In such a case, the original code becomes
`cat AC BAT0 ucsi-source-psy-USBC000:001/status` (path prefix omitted). It fails to run because you cannot `cat` a directory.


ethancedwards8:
Perfect. Thank you!

Issue #101, opened 2021-04-21
Enchancement: [Disable gray background]


coodyme:
### Disable gray background
A way to disable gray background or choose terminal background color

ethancedwards8:
> ### Disable gray background
> 
> A way to disable gray background or choose terminal background color

Sure. Feel free to send in a PR.

21st-centuryman:
I'm not sure how this would be implemented. Could it be possible to add a transparent background such as the one by powerlevel10k? Or would that mean we need to rewrite for this functionality to even be possible?

ethancedwards8:
I'm not sure how other themes implement this, or how to implement it in the first place. Looking at other themes would probably be a good place to start.

PR #299, opened 2024-10-08
Add transparent background to the powerline

Heyian:
I've tried to create the least possible if statements to achieve this. This should solve #101 

I've added the 2 following options:
1. @dracula-transparent-powerline-bg -> Default to false, set background transparency on the powerline
2. @dracula-inverse-divider -> I had to change the divider for the window panes because I couldn't get the foreground color to be transparent, only the backgroud. So I've set this inversed divider to be used if *transparent-powerline-bg* is set to *true*. I've added it to the options so the user could use another divider if he had decided to change the *show-left-sep*

I've added some new variables :
1. bg_color -> set the backgroud color
2. window_sep_fg -> set the foreground color of the windows separator (it is necessary to change it if we use the inversed divider)
3. window_sep_bg -> set the background of the windows separator, same reasons as *window_sep_fg*
4. window_sep -> sets the divider used between the windows.

Those variables and their placement in the code allows no other if statement than setting those variables. I've tried to name the variables with something similar that was already used.

I've checked the powerline with *transparent-powerline-bg* set as false or without using it in *tmux.conf* and I can't find any differences from the original code.

This change doesn't seem to break anything. I haven't touched the right side of the bar since it seemed to be filled. I don't use much plugins though but I've tested it with **ssh-session**, **git** and **playerctl** and I can't find anything that breaks.

### DIFF
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -54,10 +54,8 @@ main()
   # Set color of background
   if $transparent_powerline_bg; then
     bg_color="default"
-    separator_bg_color="default"
   else
     bg_color=${gray}
-    separator_bg_color=${white}
   fi
 
   # Handle left icon configuration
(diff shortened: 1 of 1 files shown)


ethancedwards8:
Can you attach screenshots? I'm not sure I'm getting the intended effect using Kitty on MacOS.

ethancedwards8:
Screenshot 2024-10-09 at 14 47 58

My only concern is this little circle in the lower left hand corner. Can you reproduce?

PR #300, opened 2024-10-09
Bug/no wifi icon on linux

Theoreticallyhugo:
there was an issue preventing the wifi label option from being shown on linux. upon fixing that i realised that whitespaces kept being stripped from tmux options, so i fixed that issue too.


ethancedwards8:
Thanks. These are nice code cleanup improvements too.

PR #303, opened 2024-10-15
mac: Fix network detection

fvincenzo:
There air wifi networks that are not compatible with AirPort. In such a case the network.sh script does not detect correctly the SSID.

The message reported is:
  "You are not associated with an AirPort network."

Fix the script detecting correctly the SSID of networks that do not support AirPort.

### DIFF
--- a/scripts/network.sh
+++ b/scripts/network.sh
@@ -24,8 +24,13 @@ get_ssid()
       ;;
 
     Darwin)
-      if networksetup -getairportnetwork en0 | cut -d ':' -f 2 | sed 's/^[[:blank:]]*//g' &> /dev/null; then
-        echo "$wifi_label$(networksetup -getairportnetwork en0 | cut -d ':' -f 2)" | sed 's/^[[:blank:]]*//g'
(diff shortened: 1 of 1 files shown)


ethancedwards8:
We actually have an existing PR https://github.com/dracula/tmux/pull/291 . Thank you thought.

fvincenzo:
Not a problem, happy to help. According to me PR #291 has a couple of problems in it's last iteration:
- Invokes ipconfig twice.
- If the network has airport support `networksetup` should be the preferred way to detect the SSID.
Where should I put these comments?

ethancedwards8:
> @ethancedwards8 Not a problem, happy to help. According to me PR #291 has a couple of problems in it's last iteration:
> 
>     * Invokes ipconfig twice.

That's a good point.
> 
>     * If the network has airport support `networksetup` should be the preferred way to detect the SSID.
>       Where should I put these comments?

I'm not sure I understand. Why is that preferable?

fvincenzo:
> I'm not sure I understand. Why is that preferable?
So that we have a fallback mechanism in place, if `networksetup` fails we fallback on ipconfig.

ethancedwards8:
Thank you for your contribution.

PR #308, opened 2024-10-21
weather.sh: Introduce fallback when wttr.in reaches call limit

fvincenzo:
When wttr.in reaches call limit returns the following:

Sorry, we are running out of queries to the weather service at the moment. Here is the weather report for the default city (just to show you what it looks like). We will get new queries as soon as possible.
You can follow https://twitter.com/igor_chubin for the updates.

Which makes weather.sh display a random error.

Make the error predictable.

### DIFF
--- a/scripts/weather.sh
+++ b/scripts/weather.sh
@@ -53,9 +53,14 @@ display_weather()
   temperature=$(echo $weather_information | rev | cut -d ' ' -f 1 | rev) # +31°C, -3°F, etc
   unicode=$(forecast_unicode $weather_condition)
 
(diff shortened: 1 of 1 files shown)


mamantoha (approved):

fvincenzo:
Hello @ethancedwards8,

Unless you have something against it, in which case I am happy to address, could you please merge this patch?

Yesterday I saw again the error.

Thanks!

fvincenzo:
@ethancedwards8 @danerwilliams Any update on this change?

ethancedwards8:
Thank you for your contribution! I apologize for the late merge.

PR #310, opened 2024-10-23
Adding GPU Use/mem/power for AMD & Intel GPUs

shawarden:
Changes unknown outputs by `scripts/gpu_*` to be the correct output for 1 or more AMD/Intel GPUs

`GPU 0%|0%  VRAM 34MB/4.0GB|1.2GB/8.0GB  GPUP 1/43W|16/156W`

### DIFF
--- a/scripts/gpu_power.sh
+++ b/scripts/gpu_power.sh
@@ -52,9 +52,15 @@ get_gpu()
   elif [[ "$gpu" == apple ]]; then
     usage="$(sudo powermetrics --samplers gpu_power -i500 -n 1 | grep 'GPU Power' | sed 's/GPU Power: \(.*\) \(.*\)/\1\2/g')"
   else
-    usage='unknown'
+    usage=$(
+      for card in /sys/class/drm/card?
+      do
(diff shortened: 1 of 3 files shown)


shawarden:
That's a great point. There should still be an unknown fallback.

I'll see if I can beg/borrow/steal an Intel laptop at work to test it on.

Though that raises an interesting question. As is, it detects multiple GPUs that share the data source structure, but having it handle something like an integrated Intel with discrete Nvidia or AMD, as many laptops have, would be handy. I have neither... but again, work may have something I can pilfer for a day.

shawarden:
I can only find devices with Intel and Matrox iGPUs, no discreet or Arc ones.

What I have show up as `Intel` or `Matrox` respectively but cannot use this workflow to collect the needed data so only using this for `Advanced` is the way to go.

Theoreticallyhugo:
awesome! the only thing left to talk about is the normalize_percent_len that you removed

shawarden:
The values returned by vram and power aren't single entry percentages (##.#GB/##GB, and ###W/###W) so it over pads slightly. Each additional gpu doubles that padding leading to almost having as much whitespace between VRAM and the values as the values themselves take up.

Theoreticallyhugo:
that is fair, but taking away the padding from the other options isnt necessarily the way to go either. do you think you can figure out a solution where amd is unpadded but nvidia and apple keep their padding?

Theoreticallyhugo:
Thank you very much for your contribution!

PR #311, opened 2024-10-24
Feature/docs

Theoreticallyhugo:
I've attempted to write docs that are easy to navigate and mention every single option that is available in our project, in order to allow for easier customisation.
the docs are split into three parts. 
- the installation, which ive split off, but not changed in any way beyond that
- the color theming, which was added in a previous pr
- the CONFIG.md which explains every configuration option besides color theming

this pr is lacking explanations on the plugins fossil and terraform, because i honestly dont know what they are doing, and i have not yet been interested in looking into them very much, as this was quite the endeavour already.
additionally ive barely changed the README.md. while i feel like the features section is lacking a concept, i also wasnt too sure how to refactor it, so ill leave that to another pr.

so, all in all, we now have docs describing all plugins and configuration options. everything is linked for easy navigation, and the plugin list is sorted alphabetically.

i was thinking about adding info on non-bash external dependencies for each plugin. please let me know what you think about that.


fvincenzo:
@Theoreticallyhugo Amazing job! I will put some time away over the weekend to review your series! 😄 I might learn something new!

ethancedwards8:
Brilliant. Thank you.

> i was thinking about adding info on non-bash external dependencies for each plugin. please let me know what you think about that.

Good idea. I wanted to get this merged right away. Feel free to send in another PR for it. 

By the way @Theoreticallyhugo I have requested that you be added as a maintainer, but still haven't heard anything back. I will follow up with @zenorocha

Theoreticallyhugo:
awesome @ethancedwards8!
I'll write a PR for that when i got time :D

Issue #318, opened 2025-01-04
Bug: Weather does not show temperature


wisscot:
### Describe the bug
When I enable weather, it shows the weather condition (like snow/sunny icon) but not the temperature. 

### To Reproduce
1. Installed zsh, omz, tmux, tpm and dracula
2. added plugins to 
```
set -g @plugin 'tmux-plugins/tpm'
set -g @plugin 'dracula/tmux'
set -g @dracula-plugins "cpu-usage ram-usage weather time"
```
3. install plugin and reload tmux config

### Expected behavior
Temperature should show up. 

### Debug
I have added some debug in  ~/.tmux/plugins/tmux/scripts/weather.sh
```
diff --git a/scripts/weather.sh b/scripts/weather.sh
index ce3969c..957c89b 100755
--- a/scripts/weather.sh
+++ b/scripts/weather.sh
@@ -48,10 +48,16 @@ display_weather()
     display_weather='&m' # for metric system
   fi
   weather_information=$(fetch_weather_information $display_weather)
+  echo "Weather Info: $weather_information" >> /tmp/weather-debug.log

   weather_condition=$(echo $weather_information | rev | cut -d ' ' -f2- | rev) # Sunny, Snow, etc
   temperature=$(echo $weather_information | rev | cut -d ' ' -f 1 | rev) # +31°C, -3°F, etc
   unicode=$(forecast_unicode $weather_condition)
+  echo "Condition: $weather_condition, Temperature: $temperature" >> /tmp/weather-debug.log

   # Mac Only variant should be transparent on Linux
   if [[ "${temperature/+/}" == *"===="* ]]; then
```

and output to /tmp/weather-debug.log
```
Weather Info: Light snow, mist -3°C
Condition: , Temperature:
```

Seems the weather info is not parsed correctly 

### System
 - OS: Ubuntu 20.04.3 LTS
- Tmux Version: 3.0a

ethancedwards8:
I am experiencing this too. I'm guessing the API has changed in some way.

ethancedwards8:
Try this? https://github.com/dracula/tmux/pull/319

wisscot:
thanks for the quick reply!  I will try that.  In the meanwhile, I made a fix too https://github.com/dracula/tmux/pull/320

wisscot:
@ethancedwards8 #319 seems not working for me, the output is
```
Weather Info: Light snow, mist -7°C
Condition: , Temperature:
```

[...]

PR #320, opened 2025-01-04
bug fix: temperature not showing for certain weather info

wisscot:
To fix issue https://github.com/dracula/tmux/issues/318

### DIFF
--- a/scripts/weather.sh
+++ b/scripts/weather.sh
@@ -49,8 +49,8 @@ display_weather()
   fi
   weather_information=$(fetch_weather_information $display_weather)
 
-  weather_condition=$(echo $weather_information | rev | cut -d ' ' -f2- | rev) # Sunny, Snow, etc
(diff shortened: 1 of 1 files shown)


Theoreticallyhugo:
Thank you for your contribution!

considering your issue, i assume that you've tested this on ubuntu?

wisscot:
yes, it works on ubuntu

Theoreticallyhugo:
testing on a fresh fedora 41, both the original code, as well as your fix is working.
if it works on macos too, i would wait for a few days for others to check and then it would be ok to be merged imo.

ethancedwards8:
Works on MacOS. Thanks!

Theoreticallyhugo:
Awesome!
Thanks a lot for your contribution @wisscot :D

wisscot:
Thank you both for the quick action!

PR #323, opened 2025-01-22
feat(kubernetes-context): Support AWS govcloud accounts

luehm:
# Fixes
- Correctly parses ARNs from EKS clusters within a govcloud AWS tenant (example: `arn:aws-us-gov:eks:us-gov-west-1`)


ethancedwards8:
I'd be happy to merge given that another person tests this. I don't really have a way to try it out.

Theoreticallyhugo:
i cant test it, but the regex change really seems to just also match something with gov, so non gov users should be entirely unaffected. therefore id consider this pr safe.

PR #324, opened 2025-01-25
Provide option to cwd that shortens the path to solely display the current directory

AstringentBeet:
I'm a big fan of the `cwd` feature, but with the amount of directories and subdirectories I have, seeing the verbosity of my paths displayed by Dracula's `cwd` can be a bit annoying, especially when I have other plugins displayed within the powerline.

#### Into this: `CurrentDirectory` 

To do so, simply enter the following within `.tmux.conf`:
`set -g @dracula-cwd-basename true`

If merged, I hope this provides some value to users working with this wonderful plugin.

### DIFF
--- a/scripts/cwd.sh
+++ b/scripts/cwd.sh
@@ -17,6 +17,14 @@ main() {
 
   # change '/home/user' to '~'
   cwd="${path/"$HOME"/'~'}"
+  
+  # check if the user wants only the current directory
+  basename_enabled=$(tmux show-option -gqv @dracula-cwd-basename)
+
+  if ["$basenmae_enabled" == "true" ]; then
+     #Extract only the last part of path
+     cwd=$(basename "cwd") 
+  fi
 
   echo "$cwd"
 }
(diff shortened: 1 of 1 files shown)


Theoreticallyhugo:
Hey @AstringentBeet, thats cool! with that in mind, i have another idea id like to propose. what if we rather add a setting of max-cwd-length. this way you could have the option to shorten the cwd to just the dir youre in, like you want, but you could also decide to give it a little more space. 
This also creates the question of, do we limit the amount of subdirs shown, or the amount of total characters?
thank you very much for your contribution, and please let me know what you think of this :D

ethancedwards8:
> Hey @AstringentBeet, thats cool! with that in mind, i have another idea id like to propose. what if we rather add a setting of max-cwd-length. this way you could have the option to shorten the cwd to just the dir youre in, like you want, but you could also decide to give it a little more space. This also creates the question of, do we limit the amount of subdirs shown, or the amount of total characters? thank you very much for your contribution, and please let me know what you think of this :D

Great idea to both of y'all. I think my preferred way would be subdir.

Theoreticallyhugo:
hey @AstringentBeet, 
i've attempted to implement a solution that would work for you whilst incorporating my suggestion and i would love to hear your feedback (and of course also yours ethan) :D

ethancedwards8:
Closing in favor of #326

PR #325, opened 2025-01-29
Default show_only_kubernetes_context to false

caalberts:
This fixes a bug on the `@dracula-kubernetes-context-label` option in `kubernetes_context` plugin.

`kubernetes_context.sh` is called with positional args as such:

```
kubernetes_context.sh $eks_hide_arn $eks_extract_account $hide_kubernetes_user $show_only_kubernetes_context $show_kubernetes_context_label
```

When `show_only_kubernetes_context` defaults to "", arg $5 `show_kubernetes_context_label`,  becomes arg $4 `show_only_kubernetes_context`. Effectively, the value of `@dracula-kubernetes-context-label` is passed to the script as `show_only_kubernetes_context` instead of `show_kubernetes_context_label` So the default value for `show_only_kubernetes_context` needs to be a non-zero string.

Here are screenshots showing the bug:

`@dracula-kubernetes-context-label` has no effect when `@dracula-show-only-kubernetes-context` is `""`.
Screenshot 2025-01-29 at 7 59 05 PM

`@dracula-kubernetes-context-label` has no effect when `@dracula-show-only-kubernetes-context` is not set.
Screenshot 2025-01-29 at 7 58 34 PM

Setting `@dracula-kubernetes-context-label` to `"true"` is  effectively the same as setting `@dracula-show-only-kubernetes-context` to `"true"`.
Screenshot 2025-01-29 at 8 06 36 PM

`@dracula-kubernetes-context-label` only takes effect when `@dracula-show-only-kubernetes-context` is explicitly set to `"false"` or other non-zero length string other than "true".
Screenshot 2025-01-29 at 7 58 02 PM

### DIFF
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -10,7 +10,7 @@ main() {
   show_krbtgt_label=$(get_tmux_option "@dracula-krbtgt-context-label" "")
   krbtgt_principal=$(get_tmux_option "@dracula-krbtgt-principal" "")
   show_kubernetes_context_label=$(get_tmux_option "@dracula-kubernetes-context-label" "")
-  show_only_kubernetes_context=$(get_tmux_option "@dracula-show-only-kubernetes-context" "")
+  show_only_kubernetes_context=$(get_tmux_option "@dracula-show-only-kubernetes-context" false)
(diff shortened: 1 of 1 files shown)


ethancedwards8:
Good catch. Thanks!

PR #326, opened 2025-02-02
feature: add max length feature to cwd

Theoreticallyhugo:
in reference to #324 this pr attempts to implement the suggestion.

tested on macos

not yet documented.

use `set -g @dracula-cwd-length "0"` to set the max length, where 0 is unrestricted.

whenever the paths length is cut, it is displayed without / or ~/ to the left of it to prevent confusion.

example:
max length is 1:
/ becomes / as usual
/Users becomes /Users as usual
/Users/username becomes ~ as usual
/Users/username/.config becomes ~/.config as usual
/Users/username/.config/tmux becomes tmux. no / to the left indicates a cut path
max length is 3:
/Users/username/.config becomes ~/.config as usual
/Users/username/.config/tmux/plugins becomes ~/.config/tmux/plugins as usual
/Users/username/.config/tmux/plugins/tmux becomes tmux. no / to the left indicates a cut path
max length is 0:
unchanged behaviour to previously

my solution seems hacky, but i couldnt think of anything else at this time of day. 
please critique or suggest a different solution.

documentation and testing on linux will follow.

### DIFF
--- a/scripts/cwd.sh
+++ b/scripts/cwd.sh
@@ -41,7 +41,7 @@ main() {
   fi
 
   cwd_max_chars="$(get_tmux_option "@dracula-cwd-max-chars" "0")"
-  if [[ "${#cwd}" -gt "$cwd_max_chars" ]]; then
+  if [[ "${cwd_max_chars}" -gt 0 && "${#cwd}" -gt "$cwd_max_chars" ]]; then
     cwd="…/…${cwd:(- cwd_max_chars)}"
   fi
 
--- a/scripts/cwd.sh
+++ b/scripts/cwd.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+current_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source $current_dir/utils.sh
+
 # return current working directory of tmux pane
 getPaneDir() {
   nextone="false"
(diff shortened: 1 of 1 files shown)


bmodotdev:
At least one edge case here. For ease of testing, I modified the script to accept limit in $1 and path in $2:
```
# HOME=/home/user
$ ./scripts/cwd.sh 3 /home/user.bak/one/two/three
~.bak/one space/two/three
```

It's a small edge case, but it could be devastating to mislead of their current directory; e.g. they could think they are safely deleting files from a specific directory.

I might try something like below:
```
# Truncation is only needed $cwd contains $HOME,
# which requires $cwd to be greater than $HOME
if (("${#HOME}" < "${#cwd}")); then
  home_length=$((${#HOME} + 1))
  [ "${HOME}/" == "${cwd::$home_length}" ] && cwd="~/${cwd:$home_length}"
fi
```

I would also add that "@dracula-cwd-length" plugin option might be misinterpreted as length of chars, not length of dirs. Perhaps there is a better name for this option like "depth". Also, speaking of length of chars, this PR doesn't really address it. You may have a dir limit of 3, but still get a huge cwd from this plugin if the dir names themselves are too long. I think it would be good to accommodate that as well.

As well, when a directory is not in the $HOME path, and it exceeds the limit, I think a specific symbol would be better than leaving it blank. Other programs like starship use ".../" or even better, an ellipsis "…/" 

And again, testing should probably be quite heavy on this, as it being incorrect could be very damaging.

scripts/cwd.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+current_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source $current_dir/utils.sh
```
bmodotdev, line 4:
```
source $current_dir/utils.sh
^-------------------^ SC1091 (info): Not following: ./utils.sh was not specified as input (see shellcheck -x).
^----------^ SC2086 (info): Double quote to prevent globbing and word splitting.
```

Theoreticallyhugo:
hi, thank you very much for your feedback!

i looked into the bad home dir substitution and think i got a fix that may run a little faster than yours (please absolutely do correct me if i'm wrong).
i also added the ellipsis you proposed. i didn't know what other programs were using for this and am happy to choose a more established solution.
and i added the max number of chars in a sensible way ... i hope.

i'd be happy if you could test the changes.

the options are now:
```bash
set -g @dracula-cwd-max-dirs "3"
set -g @dracula-cwd-max-chars "10"
```
as usual, the respective options are ignored when set to `"0"`

ethancedwards8:
Awesome, thanks!

Issue #267, opened 2024-05-28
Coordination showing when location set to false Bug


Edvid:
### Describe the bug
Sometimes, the weather part of the plugin displays my coordinates, despite having `set -g @dracula-show-location falsse`

### To Reproduce
1. use `set -g @dracula-plugins "hostname window cpu-usage gpu-usage ram-usage time weather"` or anything else using weather
2. `set -g @dracula-show-location false`
3. Go throughout your day using tmux (It is rare. I have seen it happen three times, and I've been using tmux for half a year)
4. See behaviour similar to mine

### Expected behavior
For this to never ever happen 

### Screenshots
Excuse the weird blur.. I just didn't want to doxx myself
![coords](https://github.com/dracula/tmux/assets/44142407/e2a6ade0-2a29-46b7-a3d6-0e47f131bdc1)

### System
 - OS: Arch Linux 
- Tmux Version: 3.4

Edvid:
Done a little digging. You make use of an API at wttr.in with curl. 
The following line dictates how you display temperature in the (sub)plugin. 
`temperature=$(echo $weather_information | rev | cut -d ' ' -f 1 | rev) # +31°C, -3°F, etc`
This takes the last word of the stdout from the fetch_weather_information() which in turn is the stdout of a curl call to wttr.in with format parameters %C and %t
The line is: `curl -sL wttr.in/${fixedlocation// /%20}\?format="%C+%t$display_weather"`. 26th in weather.sh

Sadly, sometimes the output from this API is:
`Unknown location; please try ~`<somecord> with <somecord> being replaced with coordinates

```
$ curl -SL wttr.in/~Eifal+Tuwer\?format="%C+%t&u"
Unknown location; please try ~not found

$ curl -SL wttr.in/~Eifel+Tower\?format="%C+%t&u"
Unknown location; please try ~49.9456277,6.5562824

$ curl -SL wttr.in/~49.9456277,6.5562824\?format="%C+%t&u"
Unknown location; please try ~49.9457915,6.5566064

$ curl -SL wttr.in/~49.9457915,6.5566064\?format="%C+%t&u"
Unknown location; please try ~49.9457915,6.5566064

```

This seems to not be an issue with how this plugin parses the input coordinates to the API. The fact that it breaks sometimes seemingly out of nowhere, and the following returned text from curl if we parse no format parameters gives me the impression that this is just their service being down sometimes.

the returned text from API without params:
```
Sorry, we are running out of queries to the weather service at the moment.
[...]

PR #327, opened 2025-02-03
refactor(weather): Fix timeout, reduce calls, apply shellcheck

bmodotdev:
## Issue

No issue submitted yet.

### General Overview
Fix Darwin timeout function duration argument. Remove unnecessary API calls to “ipinfo.io”. Confirm to Google Shell style guide and appease shellcheck.

The Darwin [timeout function](https://github.com/dracula/tmux/tree/dd1a7ab5a6e462a26723f63f968d79a4c2081510) makes use of variable `duration` which does not exist.

One execution of the `./scripts/weather.sh` makes 4 remote API calls, 2 calls to the weather service endpoint [wttr.in](https://wttr.in) and 2 calls to the geodata endpoint [ipinfo.io](https://ipinfo.io) to fetch city and region. We should make use of the first weather call for both error checking and data gathering. The calls to the geodata endpoint are not necessary as the weather endpoint provides this functionality. This change results in a single API call each script execution.

Exec's of awk, grep, and xargs should not be necessary, as bash parameter expansion allows us to extract data needed.

General style changes to conform to [Google Shell style guide](https://google.github.io/styleguide/shellguide.html), add function argument descriptions, and changes to appease shellcheck. Reduce scope of globals to local function variables.

[Shellcheck](https://github.com/koalaman/shellcheck) output
```
$ shellcheck --severity style scripts/weather*

perl -e 'alarm shift; exec @ARGV' "$duration" "$@"
^-------^ SC2154 (warning): duration is referenced but not assigned.


In scripts/weather.sh line 18:
if $location && [[ ! -z "$fixedlocation" ]]; then
^-- SC2236 (style): Use -n instead of ! -z.


In scripts/weather.sh line 33:
api_response=$(curl -sL wttr.in/${fixedlocation// /%20}\?format="%C+%t$display_weather")
^---------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
api_response=$(curl -sL wttr.in/"${fixedlocation// /%20}"\?format="%C+%t$display_weather")


In scripts/weather.sh line 38:
echo $api_response
^-----------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
echo "$api_response"


In scripts/weather.sh line 54:
unicode=$(forecast_unicode $weather_condition)
^----------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
unicode=$(forecast_unicode "$weather_condition")


In scripts/weather.sh line 71:
weather_condition=$(echo $weather_condition | awk '{print tolower($0)}')
^----------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
weather_condition=$(echo "$weather_condition" | awk '{print tolower($0)}')


In scripts/weather_wrapper.sh line 21:
if [ "$(expr ${TIME_LAST} + ${RUN_EACH})" -lt "${TIME_NOW}" ]; then
^--^ SC2003 (style): expr is antiquated. Consider rewriting this using $((..)), ${} or [[ ]].
^----------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
if [ "$(expr "${TIME_LAST}" + ${RUN_EACH})" -lt "${TIME_NOW}" ]; then


In scripts/weather_wrapper.sh line 23:
$current_dir/weather.sh $fahrenheit $location "$fixedlocation" > "${DATAFILE}"
^----------^ SC2086 (info): Double quote to prevent globbing and word splitting.
^---------^ SC2086 (info): Double quote to prevent globbing and word splitting.
^-------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
"$current_dir"/weather.sh "$fahrenheit" "$location" "$fixedlocation" > "${DATAFILE}"

For more information:
https://www.shellcheck.net/wiki/SC2154 -- duration is referenced but not as...
https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
https://www.shellcheck.net/wiki/SC2003 -- expr is antiquated. Consider rewr...
```

## Testing

Tested in Ubuntu as below:
```
$ ./scripts/weather.sh
☀ 70°F Houston, Texas

$ ./scripts/weather.sh true true
☀ 70°F Houston, Texas

$ ./scripts/weather.sh true false
☀ 70°F

$ ./scripts/weather.sh false false
☀ 21°C

$ ./scripts/weather.sh false true
☀ 21°C Houston, Texas

$ ./scripts/weather.sh true true "Los Angeles, California"
☀ 66°F Los Angeles, California

$ ./scripts/weather.sh true false "Los Angeles, California"
☀ 66°F

$ ./scripts/weather.sh false true "Los Angeles, California"
☀ 19°C Los Angeles, California

$ ./scripts/weather.sh "" "" "Los Angeles, California" ;echo
☀ 66°F Los Angeles, California
```
 
For the wrapper script, I manually adjusted `$INTERVAL` to 5, and added an `printf "updating cache "` to the interval condition, then:
```
$ /bin/rm /tmp/.dracula-tmux-*; ./scripts/weather_wrapper.sh true false
updating cache ☀ 70°F

$ /bin/rm /tmp/.dracula-tmux-*; ./scripts/weather_wrapper.sh false true "Los Angeles, California"
updating cache ☀ 19°C Los Angeles, California

$ /bin/rm /tmp/.dracula-tmux-*; ./scripts/weather_wrapper.sh true true "Los Angeles, California"; echo; ./scripts/weather_wrapper.sh 
updating cache ☀ 66°F Los Angeles, California
☀ 66°F Los Angeles, California

$ /bin/rm /tmp/.dracula-tmux-*; ./scripts/weather_wrapper.sh false false "Los Angeles, California"; echo; ./scripts/weather_wrapper.sh
updating cache ☀ 19°C
☀ 19°C
```

I will test on OS X tomorrow at the office.

### DIFF
--- a/scripts/weather.sh
+++ b/scripts/weather.sh
@@ -2,96 +2,110 @@
 # setting the locale, some users have issues with different locales, this forces the correct one
 export LC_ALL=C.utf8
 
-fahrenheit=$1
-location=$2
-fixedlocation=$3
+API_URL="https://wttr.in"
+DELIM=":"
 
 # emulate timeout command from bash - timeout is not available by default on OSX
 if [ "$(uname)" == "Darwin" ]; then
-  timeout() {
-      perl -e 'alarm shift; exec @ARGV' "$duration" "$@"
+  function timeout() {
+    local _duration
+    _duration="${1:-1}"
+    command -p perl -e 'alarm shift; exec @ARGV' "$_duration" "$@"
   }
 fi
 
-display_location()
-{
-  if $location && [[ ! -z "$fixedlocation" ]]; then
-    echo " $fixedlocation"
-  elif $location; then
-    city=$(curl -s https://ipinfo.io/city 2> /dev/null)
-    region=$(curl -s https://ipinfo.io/region 2> /dev/null)
-    echo " $city, $region"
+# Fetch weather information from remote API
+# Globals:
+#   DELIM
+# Arguments:
+#   show fahrenheit, either "true" or "false"
+#   optional fixed location to query weather data about
+function fetch_weather_information() {
+  local _show_fahrenheit _location _unit
+  _show_fahrenheit="$1"
+  _location="$2"
+
+  if "$_show_fahrenheit"; then
+    _unit="u"
   else
-    echo ''
+    _unit="m"
+  fi
+
+  command -p curl -sL "${API_URL}/${_location// /%20}?format=%C${DELIM}%t${DELIM}%l&${_unit}"
+}
+
+# Format raw weather information from API
+# Globals:
+#   DELIM
+# Arguments:
+#   The raw weather data as returned by "fetch_weather_information()"
+#   show location, either "true" or "false"
+function format_weather_info() {
+  local _raw _show_location
+  _raw="$1"
+  _show_location="$2"
+
+  local _weather _temp _location
+  _weather="${_raw%%"${DELIM}"*}"  # slice temp and location to get weather
+  _temp="${_raw#*"${DELIM}"}"      # slice weather to get temp and location
+  _temp="${_temp%%"${DELIM}"*}"    # slice location to get temp
+  _temp="${_temp/+/}"              # slice "+" from "+74°F"
+  _location="${_raw##*"${DELIM}"}" # slice weather and temp to get location
+  [ "${_location//[^,]/}" == ",," ] &&
+    _location="${_location%,*}" # slice country if it exists
+
+  case "${_weather,,}" in
+  'snow')
+    _weather='❄'
+    ;;
+  'rain' | 'shower')
+    _weather='☂'
+    ;;
+  'overcast' | 'cloud')
+    _weather='☁'
+    ;;
+  'na')
+    _weather=''
+    ;;
+  *)
+    _weather='☀'
+    ;;
+  esac
+
+  if "$_show_location"; then
+    printf '%s %s %s' "$_weather" "$_temp" "$_location"
+  else
+    printf '%s %s' "$_weather" "$_temp"
   fi
 }
 
-fetch_weather_information()
-{
-  display_weather=$1
-  # it gets the weather condition textual name (%C), and the temperature (%t)
-  api_response=$(curl -sL wttr.in/${fixedlocation// /%20}\?format="%C+%t$display_weather")
+# Display weather, temperature, and location
+# Globals
+#   none
+# Arguments
+#   show fahrenheit, either "true" (default) or "false"
+#   show location, either "true" (default) or "false"
+#   optional fixed location to query data about, e.g. "Houston, Texas"
+function main() {
+  local _show_fahrenheit _show_location _location
+  _show_fahrenheit="${1:-true}"
+  _show_location="${2:-true}"
+  _location="$3"
 
-  if [[ $api_response = "Unknown location;"* ]]; then
-    echo "Unknown location error"
-  else
-    echo $api_response
-  fi
-}
-
-#get weather display
-display_weather()
-{
-  if $fahrenheit; then
-    display_weather='&u' # for USA system
-  else
-    display_weather='&m' # for metric system
-  fi
-  weather_information=$(fetch_weather_information $display_weather)
-
-  weather_condition=$(echo "$weather_information" | awk -F' -?[0-9]' '{print $1}' | xargs) # Extract condition before temperature, e.g. Sunny, Snow, etc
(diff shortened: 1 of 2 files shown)


bmodotdev:
One thing that wasn't clear to me was the error checking [here](https://github.com/dracula/tmux/blob/dd1a7ab5a6e462a26723f63f968d79a4c2081510/scripts/weather.sh#L56). I tried finding an example output of the error "====" but couldn't find any in the commit, PR, or in upstream wttr.in project. It's also unclear to me why the [second conditional](https://github.com/dracula/tmux/blob/dd1a7ab5a6e462a26723f63f968d79a4c2081510/scripts/weather.sh#L61) checks for the string "error" when the first conditional sets it; i.e. we could simply return at finding the first error, unless we actually expect the string "error" to appear in addition to "====".

If you can advise on sample error output, I'd be happy to make it more robust and propagate the error message up.

Also, please feel free to let me know if these types of refactors are welcomed or not. I'd be happy to continue with any other scripts or understand if you want to focus on fixes/features only.

Thank you for the cool project!

bmodotdev:
Ahh, I found #267, which shows the error message as:
```
Sorry, we are running out of queries to the weather service at the moment.
Here is the weather report for the default city (just to show you what it looks like).
We will get new queries as soon as possible.
You can follow https://twitter.com/igor_chubin for the updates.
======================================================================================
```

So I suppose the check against "===" is to match this error message. I too would not be confident in exactly which string to match against as I don't know if that string is guaranteed not to change. But "===" is probably not the most obvious method for the reader, nor the most guaranteed string of the message.

I also copied the curl commands, added the include flag, and we see the output has changed since reported in #267 
```
$ curl -isL wttr.in/~Eifal+Tuwer\?format="%C+%t&u"
HTTP/1.1 404 Not Found
Access-Control-Allow-Origin: *
Content-Length: 25
Content-Type: text/plain; charset=utf-u8
Date: Mon, 03 Feb 2025 04:24:56 GMT

Patchy rain nearby +63°F
```

I think the appropriate thing to do would be to pivot on the status code, especially since this results in leaking user PI. We can see the unknown error is handled by a 404, and surely the status code makes a better contract.

This lead me to also search the upstream project for the [capacity text](https://github.com/chubin/wttr.in/blob/b8a9d7743205bdc227f2a88b026a6a629f0803ea/lib/translations.py#L455), which lead to the [server implementation](https://github.com/chubin/wttr.in/blob/b8a9d7743205bdc227f2a88b026a6a629f0803ea/lib/wttr_srv.py#L396).

Given this is a potential PI leak, I suggest using `curl --fail` to drop body and emit a 22 status code, which can allow the consumer to pivot easily. I don't see a reason to keep the body besides debugging.

bmodotdev:
With status code check:
```
# master `dd1a7ab`
$ ./scripts/weather.sh true true "Eifal+Tuwer"
☂  63°F Eifal+Tuwer

# PR
$ ./scripts/weather.sh true true "Eifal+Tuwer"
Unknown Location
```

With this change, I believe we can close #267, which was previously patched via #268 , but appears the API has since changed.

And I believe #308 was a patch to handle the 503 capacity error message, though I am still confused by the [comment](https://github.com/dracula/tmux/pull/308/files#diff-d5862b20c36ba74d97e26569cc0b893f319456404a1b28b5eb49ed2a9b4bfc29R56) that was added.

Theoreticallyhugo:
hi @bmodotdev, you're truly doing gods work. I've been planning on refactoring, but never got around to do it, so I'm very happy you're helping us. 
please feel free to continue with refactors like this!
your work is much appreciated :D

bmodotdev:
Thank you!

I just ran this on OSX, and got an error regarding locale:
```
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
    LC_ALL = "C.utf8"
[... etc ...]
are supported and installed on your system.
perl: warning: Falling back to a fallback locale ("en_US.UTF-8")
```

I thought `LC_ALL=C.utf8` was a strange value, but did not want to touch it without reason or knowing more.  Checking the rest of the project:
```
$ git grep -hP 'LC_ALL=\S+'|sort|uniq -c
1 export LC_ALL=C.utf8
25 export LC_ALL=en_US.UTF-8
```
Looks like this is the only script that uses that value, so I've updated it.

I also received an error regarding lowercase parameter expansion. OSX is stuck on Bash3.2 because they refused to accept the change from GPLv2 to GPLv3. I've updated it to use `tr` instead.

I believe this is PR ready for review now.

ethancedwards8 (approved):
This is great stuff. Will test then merge. Thank you so much for everything!

PR #333, opened 2025-04-01
fix sys-temp being ignored

tgalal:
The loop parsing plugins skips over sys-temp due to a slight misplacement.


ethancedwards8 (approved):
Nice catch. Thanks.

PR #338, opened 2025-04-25
Cache ip address in network-public-ip plugin

luftaquila:
Apply caching to the public IP just like the weather plugin, to prevent curl every few seconds.


luftaquila:
Sure, I don't really care about that. I just updated the code as you suggested.

Theoreticallyhugo:
thank you very much for your contribution!

PR #340, opened 2025-05-16
Fix tmux_ram_info bug

### LINKED ISSUE #321: Bug: [Excessive RAM usage reported by `tmux_ram_info.sh` plugin]
### Describe the bug
The `tmux_ram_info.sh` plugin shows an excessive RAM usage when nvim is open that does not match the actual usage.

image
Which does not match the ram that my system is using.

image

### To Reproduce
1. Go to 'tmux'
2. Run the `tmux_ram_info.sh` script
3. Observe the reported RAM usage

Note: The plugin seems to work fine, but when using `nvim` with `lazyvim`, it does not count the processes correctly.

### Expected behavior
The script should report accurate RAM usage.

### System
- OS: Arch Linux
- Tmux Version: Latest

### Possible solution

I found a possible solution but I would like to discuss it before making a PR. The original code on line 74 does this:

```sh
total_mem_kb="$(ps -o rss= -p "$pids" | paste -sd+ - | bc)"
```

So I changed it to this:

```sh
total_mem_kb="$(ps -o rss= -p "$pids" | awk '!seen[$0]++' | paste -sd+ | bc)"
```

Which shows me a better representation:
image

> NOTE: I adjusted the rounding to better display the values, but in my solution, the rounding will remain the same.

When I run the command `pstree -p "$pid"` with the tmux PID, it shows a tree with multiple processes having different PIDs:

image

And when extracting their sizes in bytes using `ps -o rss= -p "$pids"` it give us this result:

```
 8968
10628
47736
 9632
69208
151140
151140
151140
151140
151140
151140
151140
151140
151140
151140
151140
221548
221548
221548
221548
221548
221548
221548
221548
221548
221548
221548
69208
69208
69208
69208
47736
47736
47736
47736
47736
47736
47736
47736
47736
11356
10268
 9960
```

However, it appears that we should not sum all of them because some PIDs are repeated. My solution was to sum only the unique values to `total_mem_kb`, but I am not sure if it's the best solution for this bug.

aaronkollasch:
Occurs with lines too long that get truncated, causing unclosed parentheses and breaking the sed regex, leading to an empty output. The `-l` option fixes this.

The `-T` option is for removing unnecessary PIDs (thread PIDs) from pstree output and may remove the duplicate PIDs seen in #321.

May fix #321

### DIFF
--- a/scripts/tmux_ram_info.sh
+++ b/scripts/tmux_ram_info.sh
@@ -67,7 +67,7 @@ get_tmux_ram_usage()
   case $(uname -s) in
     Linux)
       if command -v pstree > /dev/null; then
-        pids="$(pstree -p "$pid" | tr -d '\n' | sed -rn -e 's/[^()]*\(([0-9]+)\)[^()]*/\1,/g' -e 's/,$//p')"
+        pids="$(pstree -Tlp "$pid" | tr -d '\n' | sed -rn -e 's/[^()]*\(([0-9]+)\)[^()]*/\1,/g' -e 's/,$//p')"
       else
         pids="$(get_cpids_linux "$pid" | tr '\n' ',')"
       fi
@@ -76,7 +76,7 @@ get_tmux_ram_usage()
 
     Darwin)
       if command -v pstree > /dev/null; then
-        pids="$(pstree "$pid" | sed -En 's/[^0-9]+([0-9]+) .*/\1/p' | tr '\n' ',')"
+        pids="$(pstree -w "$pid" | sed -En 's/[^0-9]+([0-9]+) .*/\1/p' | tr '\n' ',')"
       else
         pids="$(get_cpids_unix "$pid" | tr '\n' ',')"
       fi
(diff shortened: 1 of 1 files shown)


ethancedwards8:
I'll review this later today when I'm at a PC. 

As a side note, I saw your profile says you're a postdoc at Harvard. I'll be starting my undergrad there in the fall. Was that where your undergrad was as well?

aaronkollasch:
Thanks @ethancedwards8, and congrats!

I did my undergrad at Northwestern and PhD at Harvard Medical School (and stayed on for about a year as a postdoc) 🙂

ethancedwards8 (approved):

ethancedwards8:
Thanks for the contribution.

And thanks! Congrats on the PhD and postdoc.

aaronkollasch:
Thanks!

Also, just checked with a neovim instance and `pstree -T` does indeed remove the excess thread pids (though I actually didn't see the double-counting bug for some reason, even with all the thread pids).

PR #341, opened 2025-05-17
Fix cpu_info bug

aaronkollasch:
Grepping the output of top for `Cpu(s)` includes a row for the `grep "Cpu(s)"` process, which sometimes ends up being selected by `tail -1`, causing a nonsensical %CPU reading (in orange in the screenshot below).

Screenshot 2025-05-16 at 2 22 58 AM

This fix uses the pattern `[C]pu(s)` to exclude the search pattern in the grep process itself, a la this stackexchange https://unix.stackexchange.com/a/74186.
Screenshot 2025-05-16 at 2 24 47 AM

It is still possible for the grep pattern to match `Cpu(s)` in the arguments for another process, but that seems much less likely.

### DIFF
--- a/scripts/cpu_info.sh
+++ b/scripts/cpu_info.sh
@@ -9,7 +9,7 @@ get_percent()
 {
   case $(uname -s) in
     Linux)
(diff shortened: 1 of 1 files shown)


ethancedwards8 (approved):

ethancedwards8:
Nice catch. Thanks.

Issue #316, opened 2024-12-19
Enhancement: Support IP2Location.io API


ip2location:
### Describe the feature you'd like
Add a support of our IP2Location.io API as an alternative to your project to query for the region and city details. Our service provided 500 queries per day for keyless query, or 30000 queries per month if you sign up our free plan. You can read more on the result returned via [our documentation](https://www.ip2location.io/ip2location-documentation).

PR #344, opened 2025-05-29
feat:spr plugin

iberniex:
This is based on the creation of spotify-player widget for the spotify-player app based on #316 

There is also a remote option but it has some issues with latency and needs the daemon to be installed but limits the refresh rate on the display widget.

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -1,6 +1,7 @@
 # [tmux](https://github.com/tmux/tmux/wiki)
 
 ## Table of Contents
+
 - [Configuration](#Configuration---up)
 - [Status bar options](#status-bar-options---up)
   - [Powerline](#powerline---up)
@@ -29,6 +30,7 @@
   - [playerctl](#playerctl---up)
   - [ram-usage](#ram-usage---up)
   - [spotify-tui](#spotify-tui---up)
+  - [spr](#spr--up)
   - [ssh-session](#ssh-session---up)
   - [synchronize-panes](#synchronize-panes---up)
   - [sys-temp](#sys-temp---up)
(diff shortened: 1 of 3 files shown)


docs/CONFIG.md
```diff
@@ -515,85 +581,147 @@ set -g @dracula-ping-rate 5
...
 ```
 
 To limit the maximum length (0 means unlimited length):
+
 ```bash
 set -g @dracula-spotify-tui-max-len 30
 ```
 
-`set -g @dracula-refresh-rate 5` affects this widget
+`set remote-g @dracula-refresh-rate 5` affects this widget
```
Theoreticallyhugo, line 636: is this change intentional?

iberniex:
is it when you do the 'R' remote command? 
Oh cause I just noticed that when you use the 'R' keybind, it displays the osascript mac player command so I'll do a slight patch on the mac-player plugin for the remote feature which was previously like this.

```bash
  # Remote Access
  if [[ "$REMOTE_ACCESS" == true ]]; then
    remoteControl "$PLAY_PAUSE_BUTTON" "$BACK_BUTTON" "$NEXT_BUTTON" "$REMOTE_APP"
  fi
```

then add a guard to prevent the commands from leaking to other commands in tmux. my bad :p .

```bash
  # Remote Access
  if [[ "$REMOTE_ACCESS" == true ]]; then
    remoteControl "$PLAY_PAUSE_BUTTON" "$BACK_BUTTON" "$NEXT_BUTTON" "$REMOTE_APP"
  else
    tmux unbind-key "$toggle_button"
    tmux unbind-key "$back_button"
    tmux unbind-key "$next_button"
  fi
```
>besides it's displaying well! formatting options like there used to be with spotify-tui would be cool, but not a must (set -g >@dracula-spotify-tui-format "%f %s %t - %a")

for the formatting options, I was finding ways to store the fields separately to the formatting options like in spotify-tui cause spotify-player doesn't provide the formatting options. The data they expose is in json but I'll figure something out.

PR #347, opened 2025-06-25
new battery status "Not charging" and nerdfont docs

Theoreticallyhugo:
now the status "Not charging" is added and considered the same as MacOS' "ACattached", since both mean that the battery is neither charging nor discharging, but power is being drawn form AC only.
also, the icons are now described in the docs. my github still can't render them, but it's good to have it there for when people check it some other way.


ethancedwards8:
Nice catch. Thanks.

PR #349, opened 2025-07-09
feat(): Show repo name status bar

lfreixial:
Show the repo name on the status bar like
`tmux | Main`

Can be enabled and disabled by having `@dracula-git-show-repo-name` in the config file. False by default.


lfreixial:
> hi, thank you very much for your pr, this is very cool! i made changes, some of which were stupid and so i immediately undid them, but i stand by moving that white space for all those who won't activate the feature. i'll leave this pr up for a day or to so that you got time to check my changes and then I'll merge. this looks good to me

Awesome thank you so much! and yeah I agree removing the extra white space is a good call.

PR #350, opened 2025-07-12
feat/add gsm support

Theoreticallyhugo:
network.sh now supports gsm modules via nmcli.

when connected via gsm, the gsm label is displayed in combination with the carrier.  
the gsm label can be set to something custom:
```bash
set -g @dracula-network-gsm-label " "
```

tested on:
 - [x] linux (x86) with gsm
 - [ ] linux (x86) without gsm
 - [x] MacOS (apple silicon) without gsm


Theoreticallyhugo:
@coderabbitai review

PR #352, opened 2025-07-16
feat/making plugin padding customizable

Theoreticallyhugo:
replace the space that would be added to the left and right of a plugin by customizable variables.
both the left and right pad variables are one space per default in order not to change the status quo and break user configs.

this enables tighter packing of plugins, by using a zero width space for padding.  
```bash
set -g @dracula-left-pad '​'  # THERE IS A ZERO WIDTH SPACE HERE
set -g @dracula-right-pad '​'  # USING '' WITHOUT A PADDING CHARACTER WON'T WORK
set -g @dracula-show-right-sep " "
set -g @dracula-show-powerline true
```
image

or fancy wide configs
```bash
set -g @dracula-show-edge-icons true
set -g @dracula-left-pad '   '
set -g @dracula-right-pad '   '
set -g @dracula-show-left-sep " "
set -g @dracula-show-right-sep " "
set -g @dracula-inverse-divider 
```
image

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -88,6 +88,18 @@ Hide empty plugins
 set -g @dracula-show-empty-plugins false
 ```
 
+Set plugin padding
+Whilst the padding is one space per default, can be whatever you want it to be, whether that's whitespace or other characters.
+**If you want to remove any padding, you need to use a zero width space!**
+
+```bash
+set -g @dracula-left-pad ' ° '
+set -g @dracula-right-pad ' ° '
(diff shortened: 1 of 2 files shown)


Theoreticallyhugo:
hi @druxorey, may i ask you to review this pr? i feel like this is something you probably know well and may like

druxorey:
Hello, sorry for the delay. I've already checked and tested the PR, and it's a great addition. I haven't found any problems, so I think you can merge it into main.

PR #353, opened 2025-07-18
feat: Hide battery widget on Desktop PCs

lfreixial:
Checks to see if there is no battery and if `@dracula-battery-hide-on-desktop"` is true it will hide the battery widget.

Useful for when a config is shared between a Laptop and a Desktop and only have the battery hidden on the Desktop

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -219,6 +219,12 @@ alternatively, if you have no battery and would like a nerdfont icon to indicate
 set -g @dracula-no-battery-label " "
 ```
 
+if you only want to hide the battery widget on a desktop pc, set the following:
+
+```bash
+set -g @dracula-battery-hide-on-desktop true
+```
+
 ### compact-alt - [up](#table-of-contents)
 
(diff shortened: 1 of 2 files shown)


ethancedwards8:
Nice catch. Thanks.

Theoreticallyhugo:
hi, we actually already had a solution for that:
```bash
set -g @dracula-no-battery-label false
```

ethancedwards8:
@Theoreticallyhugo I apologize. I missed that.

Feel free to revert and add the appropriate documentation.

Also @lfreixial , I apologize to you as well.

lfreixial:
@Theoreticallyhugo I saw that one but I wasn't 100% sure what it was meant to do as when I tried using it the battery still appeared and when I manually run the `battery.sh` script I get this output. 
```bash
ls: cannot access '/sys/class/power_supply/*': No such file or directory
♥
```

Perhaps there is bug somewhere causing it to not disappear with  `@dracula-no-battery-label`. 
I can try and create a new PR to remove the new way and fix the bug.

PR #354, opened 2025-07-19
fix: Hide when on a desktop

lfreixial:
Moved the check of the battery percentage to the top as it was always hitting the $bat_stat check making it to never hide the label and always display a <3

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -219,13 +219,6 @@ alternatively, if you have no battery and would like a nerdfont icon to indicate
 set -g @dracula-no-battery-label " "
 ```
 
-if you only want to hide the battery widget on a desktop pc, set the following:
-
-```bash
-# default is false
-set -g @dracula-battery-hide-on-desktop true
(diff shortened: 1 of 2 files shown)


Theoreticallyhugo:
hi, now i must apologize i think because you're putting in the effort and i'm asking you to do more :sweat_smile: 

would you mind testing #348 where i'm making major changes to the battery script?
i would recommend setting 
```bash
# setting no label
set -g @dracula-battery-label false
set -g @dracula-no-battery-label false
```
to get no output.  

though to be fair that would kinda limit reusability of the same config on different machines in the sense that you either have a label which will prevent the plugin from disappearing on desktop, or you remove the label and thus can't have it on the laptop either.  
i'm open to discussing whether this is an issue we need to fix, or whether thats an ok restriction

Theoreticallyhugo:
thank you very much for your help!
i'll manage everything when i got time and hopefully remember to ping you so that you know when the master branch is ready :D

PR #343, opened 2025-05-29
Mac player

iberniex:
Sorry for the delay, this is based on the enhancment pull request #336 

The following things were done based on the tasks assigned to myself:
 - [x] nativePlayer search 
      - [x] Spotify
      - [x] Apple Music - Music
      - Tidal - not natively supported for applescript
 - [x] browserPlayer (The active tab has to be youtube or spotify)
      - [x] Google Chrome
      - [x] Safari
      - Firefox ( not applescript supported )
      - Chromium and Firefox variants (arc, zed) - (not applescript supported)

Contains a remote feature but the apps only supported are:
- spotify
- music (apple music)

PR #356, opened 2025-07-23
Added scrolling text to Mac player module and cleaned up functions

nutt-y:
Initially, I did some work that added a small scrolling text effect to the `playerctl` plugin last year (#273). I revisited the plugin and decided to do some clean up on it. I decided to expose some more parameters to users such as defining the width of the plugin and also the scrolling speed.

Additionally, I love the new Macos player plugin (#343) and decided to also add a similar scrolling effect for users. I exposed the scrolling speed as a parameter for users that wish to have this option on. I believe that since both of these plugins have very similar functions, there should be a common function that can be used to get the scrolling characters on screen. This function was added to the `utils.sh` file. 

Let me know if you have any questions or if there's something that needs to be fixed up!

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -25,6 +25,7 @@
   - [mac-player](#mac-player---up)
   - [mpc](#mpc---up)
   - [network](#network---up)
+  - [network-public-ip](#network-public-ip---up)
   - [network-bandwidth](#network-bandwidth---up)
   - [network-ping](#network-ping---up)
   - [network-vpn](#network-vpn---up)
@@ -91,14 +92,17 @@ set -g @dracula-show-empty-plugins false
 
 Set plugin padding
 Whilst the padding is one space per default, can be whatever you want it to be, whether that's whitespace or other characters.
-**If you want to remove any padding, you need to use a zero width space!**
 
 ```bash
 set -g @dracula-left-pad ' ° '
 set -g @dracula-right-pad ' ° '
-# no padding with zero width space
-set -g @dracula-left-pad '​'
-set -g @dracula-right-pad '​'
+```
+
+If you want to remove any padding, you can use `false` as a value.
+
+```bash
+set -g @dracula-left-pad false
+set -g @dracula-right-pad false
 ```
 
 ### Powerline - [up](#table-of-contents)
@@ -217,8 +221,6 @@ these settings will introduce the following icons:
 - power is being drawn from AC, but the battery is neither charging nor discharging: ``
 - we were able to determine that the battery is charging/ discharging, but something about the percentage went wrong: `󰂃`
 - we don't know the status of the battery: ``
-
-
 
 if you have no battery and would like the widget to hide in that case, set the following:
 
@@ -541,7 +543,6 @@ set -g @dracula-kubernetes-eks-extract-account true
 This script retrieves and displays continuous glucose monitoring (CGM) data from the LibreView API.
 It caches the data to minimize API requests and displays the latest glucose level along with a trend indicator in a Tmux status bar.
 
-
 ### mac-player - [up](#table-of-contents)
 
 This widget and script displays music information provided by the native macOS players.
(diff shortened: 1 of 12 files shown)


scripts/mac-player.sh
```diff
@@ -128,20 +127,18 @@ function trackStatus() {
...
-case "$active_player" in
-		"not running") echo "not running" ;;
-		"stopped") echo "stopped" ;;
-    "can't encode") echo "unable to encode" ;;
-    "Not Supported") echo "not supported" ;;
-
-		*) echo "$active_player" ;;
-	esac
+  *) echo "$active_player" ;;
+  esac
```
LinkUpGames, line 137: Sorry about the formatting changes, I didn't realize my lsp picked up on a couple of stylistic differences. I can go back and make sure these don't change

Theoreticallyhugo:
i didn't think about the fact that merging @iberniex plugin improvements would mess with yours. i'm sorry for the merge conflicts.

iberniex:
Let me try it out :)

LinkUpGames:
All good! I'll fix the conflicts later today

scripts/mac-player.sh
```diff
@@ -227,12 +254,20 @@ main() {
...
 
-	cat "$cache_file"
+  # Allow scrolling
+  local str=$(cat "$cache_file")
+  if [ "$SCROLL" = true ] && [ "${#str}" -ge $MAX_LENGTH ]; then
+    scroll "$str" "$MAX_LENGTH" "$SCROLL_SPEED"
+  else
+    echo "$str"
+  fi
 }
```
iberniex, line 271: so this is good but you never really update the cache if you split both logics aside from each other. just handle the `[ "$SCROLL" = true ] && [ "${#str}" -ge $MAX_LENGTH ];` condition before hand in the cache handler above so that it also factors in the cached instance in temp

iberniex (changes requested):
This feature is great! just address the condition issue for the cache_file and we're good. I was looking for a better way to make it fluid between renders cause after some time it disappears but it's good as it is for now.


oh I almost forgot so for the scrolling feature you need to wrap the boolean values in strings `[ "$SCROLL" = "true"]`  and `SCROLL=$(get_tmux_option "@dracula-mac-player-scroll" 'false')` cause in bash everything is a string. It something I found out recently when I did the adjustments.

scripts/mac-player.sh
LinkUpGames: I fixed it and also made sure the bools were strings; didn't realize that was a one of bash's quirks

scripts/mac-player.sh
```diff
@@ -227,12 +254,22 @@ main() {
...
 
-	cat "$cache_file"
+  # Allow scrolling
+  local str=$(cat "$cache_file")
+  if [ "$SCROLL" = "true" ]; then
+    scroll "$str" "$MAX_LENGTH" "$SCROLL_SPEED"
+  else
+    echo "$str"
+  fi
 }
```
iberniex, line 273:
```bash
# handle cache separately from the scrolling feature
  if [ ! -f "$cache_file" ] || [ $(($(date +%s) - $(stat -f%c "$cache_file"))) -ge "$RATE" ]; then
    local full_track
    full_track=$(trackStatus "$PAUSE_ICON" "$PLAY_ICON")

    echo "$full_track" > "$cache_file"
  fi


  # Allow scrolling and if not default to the string stripping concatenation thingy from length
  local final_str
  final_str=$(cat "$cache_file")
  if [ "$SCROLL" = "true" ] && [ "${#final_str}" -ge "$MAX_LENGTH" ]; then
    scroll "$final_str" "$MAX_LENGTH" "$SCROLL_SPEED"
  else
    sliceTrack "$final_str" "$MAX_LENGTH"
  fi
```

LinkUpGames:
Thank you for the pointers! Let me know if that looks better :)

iberniex (approved):
This is great! thanks for the feature I can approve of this :)
back you to @Theoreticallyhugo

scripts/mac-player.sh
Theoreticallyhugo: I know that it's not part of this PR, but this logic is crashing the entire script for me when spotify is not installed. If anyone could look into this, it would be greatly appreciated.
error:
```
834:839: syntax error: Expected “then”, etc. but found identifier. (-2741)
```

PR #360, opened 2025-09-29
Synchronize Panes: new feature `auto-hide`

webdavis:
Ello,

This PR does the following

- Allows folks to automatically hide the sync label when it's off via `dracula-synchronize-panes-auto-hide`.
- Allows them to also set a custom refresh rate for it via `@dracula-synchronize-panes-refresh-rate`
  - Note: this won't effect the global refresh rate managed by `@dracula-refresh-rate`

Here's how to configure it in practice:

```tmux
set-option -g @dracula-synchronize-panes-label "Sync"
set-option -g @dracula-synchronize-panes-auto-hide true  # Defaults to 'false'
set-option -g @dracula-synchronize-panes-refresh-rate "0.5"  # Defaults to no value
set-option -g @dracula-plugins "ssh-session cpu-arch synchronize-panes"
```

I'm already using this and I haven't noticed any issues (yet). Let me know what you think or any changes I need to make to get this merged.

Cheers!


ethancedwards8:
Thank you for your contribution!

PR #361, opened 2025-10-10
change spr.sh to display playback info instead of back button

j4ck4L0up3:
Ran into a bug where all I saw on the powerline was "R", fixed it.


ethancedwards8 (approved):
thanks

PR #365, opened 2025-11-27
Title music fix in spr.sh

viiniciusbonfim:
Title music was displaying back track button instead the real title music when using spr plugin.


ethancedwards8:
Closing in favor of https://github.com/dracula/tmux/pull/361 since it came first

PR #369, opened 2025-12-09
FEAT: (NVIDIA) support multiple gpus status individually

dev-onejun:
Hello!

I want to suggest to show gpus status independently from the experience that each gpu runs different jobs, if there are multiple gpus. (especially when users share the machine)

Please refer the attached image below.

Screenshot from 2025-12-09 12-11-24

It still support a one gpu as well. 

Screenshot from 2025-12-09 12-23-38

Any additional suggestions are welcome.

Thank you.

### DIFF
--- a/scripts/gpu_ram_info.sh
+++ b/scripts/gpu_ram_info.sh
@@ -46,10 +46,9 @@ get_gpu()
     echo $usage
     exit 0
     else
-      # to add finer grained info
       used_accuracy=$(get_tmux_option "@dracula-gpu-vram-used-accuracy" "d")
       total_accuracy=$(get_tmux_option "@dracula-gpu-vram-total-accuracy" "d")
-      usage=$(nvidia-smi --query-gpu=memory.used,memory.total --format=csv,noheader,nounits | awk -F ', *' '{ printf("|%dGB/%dGB", $0 / 1024, $2 / 1024) }' && echo "|")
+      usage=$(nvidia-smi --query-gpu=memory.used,memory.total --format=csv,noheader,nounits | awk -F ", *" "{ printf(\"|%${used_accuracy}GB/%${total_accuracy}GB\", \$0 / 1024, \$2 / 1024) }" && echo "|")
     fi
   elif [[ "$gpu" == Advanced ]]; then
     usage="$(


scripts/gpu_power.sh
```diff
@@ -44,9 +44,9 @@ get_gpu()
   gpu_power_percent=$(get_tmux_option "@dracula-gpu-power-percent" false)
   if [[ "$gpu" == NVIDIA ]]; then
     if $gpu_power_percent; then
-      usage=$(nvidia-smi --query-gpu=power.draw,power.limit --format=csv,noheader,nounits | awk '{ draw += $0; max +=$2 } END { printf("%d%%\n", draw / max * 100) }')
+      usage=$(echo "Power" && nvidia-smi --query-gpu=power.draw,power.limit --format=csv,noheader,nounits | awk -F ', *' '{ printf("|%d%%", $0 / $2 * 100) }' && echo "|")
   else
-      usage=$(nvidia-smi --query-gpu=power.draw,power.limit --format=csv,noheader,nounits | awk '{ draw += $0; max +=$2 } END { printf("%dW/%dW\n", draw, max) }')
+    usage=$(echo "Power" && nvidia-smi --query-gpu=power.draw,power.limit --format=csv,noheader,nounits | awk -F ', *' '{ printf("|%dW/%dW", $1, $2) }' && echo "|")
```
Theoreticallyhugo: same as above:
```suggestion
    usage=$(nvidia-smi --query-gpu=power.draw,power.limit --format=csv,noheader,nounits | awk -F ', *' '{ printf("|%dW/%dW", $1, $2) }' && echo "|")
```

scripts/gpu_ram_info.sh
```diff
@@ -42,14 +42,14 @@ get_gpu()
   gpu_vram_percent=$(get_tmux_option "@dracula-gpu-vram-percent" false)
   if [[ "$gpu" == NVIDIA ]]; then
     if $gpu_vram_percent; then
-      usage=$(nvidia-smi --query-gpu=memory.used,memory.total --format=csv,noheader,nounits | awk '{ used += $0; total +=$2 } END { printf("%d%%\n", used / total * 100 ) }')
+      usage=$(nvidia-smi --query-gpu=memory.used,memory.total --format=csv,noheader,nounits | awk -F ', *' '{ printf("|%d%%", $0 / $2 * 100) }' && echo "|")
     normalize_percent_len $usage
```
Theoreticallyhugo: This too should be an echo because of this:

Image

```suggestion
    echo $usage
```

PR #374, opened 2025-12-17
Add documentation for network-public-ip plugin

luftaquila:
I've found out that the documentation for the `network-public-ip` plugin was missing.

This PR adds the docs of it, and add a missing space between the label and the IP value of it's output.

Thank you!

### DIFF
--- a/scripts/network-public-ip.sh
+++ b/scripts/network-public-ip.sh
@@ -20,7 +20,7 @@ main() {
     ip=$(curl -s "$IP_SERVER")
 
     network_public_ip_label=$(get_tmux_option "@dracula-network-public-ip-label" "IP")
-    echo "$network_public_ip_label $ip" > "${DATAFILE}"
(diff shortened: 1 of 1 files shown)


Theoreticallyhugo:
Thank you very much for your contribution!
I will gladly merge the docs, but the lack of a space is actually intentional, as it is supposed to be up to the users config how much white space there is.

docs/CONFIG.md
```diff
@@ -645,6 +646,16 @@ Known issues:
 
 - If for some reason `iw` is only in the path for root and not the normal user, wifi connections will be considered ethernet connections.
 
+### network-public-ip - [up](#table-of-contents)
+
+This widget displays the public IP address you're using.
```
Theoreticallyhugo: I'd suggest adding info on how the public ip is aquired, as this is an external request to a third party service.
```suggestion
This widget displays the public IP address you're using, by querying the public service `ifconfig.me`.
```

PR #388, opened 2026-04-16
Fix: macos version detection

fvincenzo:
Macos 26.4.1 broke the logic in network.sh for the detection of the WiFi network name.

Fix it through string manipulation.


fvincenzo:
@ethancedwards8 @Theoreticallyhugo Is there anything I can do to improve this patch? Or is it ready to be merged?

PR #395, opened 2026-06-27
scripts/dracula.sh: add 'fill' to message-style for tmux 3.7+

gilbertwong96:
Add `fill` to Dracula's `message-style` so the message line is fully
covered under tmux 3.7+, which now overlays messages on the status
line instead of replacing it.

## Problem

image

On tmux 3.7+, renaming a window (`prefix ,`) shows the adjacent window's name through the prompt:

image

## Repro

1. Open two windows in tmux 3.7+
2. Press `prefix ,` to rename a window
3. Type — observe the adjacent window's name showing to the right of the prompt

## Fix

Add `fill=${gray}` alongside the existing `bg=${gray},fg=${white}`.

`fill=` was added to the style parser in tmux 3.0a, so the change is safe across all currently-supported tmux versions.

## AI assistance

Drafted with [pi](https://github.com/earendil-works/pi-coding-agent) (MiniMax-M3). I reproduced the issue locally, validated the fix on tmux 3.7+, and reviewed the changes.

### DIFF
--- a/scripts/dracula.sh
+++ b/scripts/dracula.sh
@@ -171,7 +171,7 @@ main() {
   tmux set-option -g pane-border-style "fg=${gray}"
 
   # message styling
-  tmux set-option -g message-style "bg=${gray},fg=${white}"
+  tmux set-option -g message-style "bg=${gray},fg=${white},fill=${gray}"
 
   # status bar
   tmux set-option -g status-style "bg=${bg_color},fg=${white}"


ethancedwards8 (changes requested):
I would appreciate visual context/evidence of the fix. Having trouble reproducing this on my machine.

Also, please clean up the PR description and the commit messages, as well as the unnecessarily verbose comment in the source code. Also, please disclose your AI usage.

ltouret:
Hello, using version 3.7b I have the same issue, using the same steps to reproduce the error.

image

With `prefix ,`:

image

PR #403, opened 2026-08-11
feat: allow configuring network-vpn interface (default tun0)

abourjac:
Allow configuring the network interface that `network-vpn` plugin checks to display vpn status.

The option is `@dracula-network-vpn-interface` (defaults to `tun0`, which is hard coded today)


ethancedwards8 (approved):
good idea but please document the option

PR #407, opened 2026-08-22
fix: use LC_ALL=C to prevent parsing errors on localized systems

moningf:
This PR fixes #406


ethancedwards8 (approved):
this lgtm but will test locally before merging!

